PR #2, opened 2021-04-01
Update pipeline operators to Operator framework

knkski:
Also adds CI/CD


charms/kfp-persistence/metadata.yaml
```diff
@@ -0,0 +1,24 @@
...
+tags: [ai, big-data, kubeflow, machine-learning, tensorflow]
+series: [kubernetes]
+resources:
+  oci-image:
+    type: oci-image
+    description: Backing OCI image
+    auto-fetch: true
+    upstream-source: gcr.io/ml-pipeline/persistenceagent:1.5.0-rc.0
+provides:
+  kfp-persistence:
```
DomFleischmann, line 19: How is it providing this? There is no sign of it in the charm code/

charms/kfp-schedwf/metadata.yaml
```diff
@@ -0,0 +1,24 @@
...
+tags: [ai, big-data, kubeflow, machine-learning, tensorflow]
+series: [kubernetes]
+resources:
+  oci-image:
+    type: oci-image
+    description: Backing OCI image
+    auto-fetch: true
+    upstream-source: gcr.io/ml-pipeline/scheduledworkflow:1.5.0-rc.0
+provides:
+  kfp-schedwf:
```
DomFleischmann, line 19: Same as `kfp-persistence`.

charms/kfp-schedwf/requirements.txt
```diff
@@ -0,0 +1,2 @@
+ops==1.1.0
+git+https://github.com/juju-solutions/resource-oci-image@1964d748022b762b9dce6e8bb7bdf12835102c72
```
DomFleischmann, line 2: Same as before, pypi.

charms/kfp-ui/requirements.txt
```diff
@@ -0,0 +1,3 @@
+ops==1.1.0
+git+https://github.com/juju-solutions/resource-oci-image@1964d748022b762b9dce6e8bb7bdf12835102c72
```
DomFleischmann, line 2: Leaving 1 comment for multiple lines in different files would be a nice GH feature...

charms/kfp-ui/src/charm.py
```diff
@@ -0,0 +1,234 @@
...
+                for relation in self.model.relations["kfp-api"]
+                for key, value in relation.data.items()
+                if isinstance(key, Unit)
+                and not key._is_our_unit
+                and value.get("service")
+            )
+            api = {
+                "service": data["service"],
+                "port": data["port"],
+            }
```
DomFleischmann, line 97: This could be done differently by using the k8s-service interface.

DomFleischmann (commented):
Mainly some things that were recurring in all the charms.

charms/kfp-api/src/charm.py
```diff
@@ -0,0 +1,364 @@
...
+                and not key._is_our_unit
+                and value.get("database")
+            )
+        except StopIteration:
+            self.model.unit.status = MaintenanceStatus(
+                "Waiting for mysql relation data"
+            )
+            return
+
+        if not ((os := self.interfaces["object-storage"]) and os.get_data()):
```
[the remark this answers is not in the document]
knkski: I took them out because the methods as-is didn't seem to map well to a generalized interface that could support either/both ends of the relationship sending data. I do however think it would be good to have helper methods like this, they'll just probably need input from this bundle as well as @johnsca's work on the load balancer interface as to how exactly they should handle more complex scenarios.

charms/kfp-persistence/metadata.yaml
knkski: Whoops, removed

charms/kfp-api/src/charm.py
```diff
@@ -0,0 +1,364 @@
...
+        config = self.model.config
+
+        try:
+            data = next(
+                value
+                for relation in self.model.relations["kfp-viz"]
+                for key, value in relation.data.items()
+                if isinstance(key, Unit) and not key._is_our_unit
+            )
+            viz = {"service": data["service"], "port": data["port"]}
```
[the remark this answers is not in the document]
knkski: Upgraded to reference https://github.com/canonical/operator-schemas/blob/master/k8s-service.yaml

charms/kfp-ui/src/charm.py
knkski: Upgraded to reference https://github.com/canonical/operator-schemas/blob/master/k8s-service.yaml

charms/kfp-api/src/charm.py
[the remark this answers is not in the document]
DomFleischmann: Yes, that makes sense. Maybe some flags like `data_available` and `data_changed` would make sense, so irregardless of what kind of endpoint it is if the charm expects data it can check those.

DomFleischmann (approved):
Code LGTM, but tests seem to have timed out.

charms/kfp-api/metadata.yaml
```diff
@@ -0,0 +1,34 @@
...
+requires:
+  mysql:
+    interface: mysql
+  object-storage:
+    interface: object-storage
+    schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/object-storage.yaml
+    versions: [v1]
+  kfp-viz:
+    interface: k8s-service
+    schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/k8s-service.yaml
```
DomFleischmann, line 27: Why is service-port type `string`

knkski: The advice I've seen that generally works well is that you should only use integers for things that you use math on. Between that and a string being required in a situation like `args: ['--port', config['port']]` (because the pod spec specifies a list of strings only, and doesn't try to convert the list), I chose a string type.

That being said, elsewhere the pod spec requires ports to be specified as integers, such as in the `ports: {..}` section. So if the first paragraph didn't convince you that it was worth using a string, I don't have any qualms about switching it.

DomFleischmann: I think we can leave it like that, not a hill I need to die on. I just wanted to make sure it was intentional.

DomFleischmann (approved):
The test failure seems the same as in all the other PRs.

Code LGTM

PR #7, opened 2021-05-26
Remove multiuser Config

DomFleischmann:
Some environment variables were enabling multiuser isolation which we aren't supporting currently on Charmed Kubeflow.

This should fix this bug: https://bugs.launchpad.net/kubeflow-bundle/+bug/1929248

More information on multiuser isolation in pipelines [here](https://www.kubeflow.org/docs/components/pipelines/multi-user/)


ktsakalozos:
LGTM +1

PR #13, opened 2021-06-10
Remove nested navigation sidebar

### LINKED ISSUE #12: Nested tabs repeated
The pipelines UI is nested with repeated tabs. This is odd. 

Screenshot from 2021-05-26 17-46-31

Low priority, but would be nice to fix.

knkski:
If pipelines are running as part of kubeflow, the frontend doesn't need to also be exposing a sidebar. Make sidebar navigation a configuration option.

Fixes #12


joedborg (approved):
LGTM

PR #14, opened 2021-06-11
Fix kfp-api extra service name

### LINKED ISSUE #6: Cannot connect to ml-pipeline service
Connecting to ml-pipeline Service from Jupyter (running in the same cluster) fails, with "connection refused". 

Turns out that the selector for ml-pipeline was not working, and so the Service was not pointing to any valid endpoints. This is confirmed in Azure Portal, where the Kubeflow cluster is deployed:

Screenshot 2021-05-04 at 5 02 38 PM

Editing the YAML file for the ml-pipeline Service seems to work around the issue temporarily:

```yaml
spec:
  selector:
    # juju-app: kfp-api (Remove this and replace with the following)
    app.kubernetes.io/name: kfp-api
```

After applying the change, the connection to ml-pipeline Service from Jupyter works.

But this is only a temporary workaround, as the YAML configuration reverts to the default (with "juju-app" as the key) after a few minutes.

knkski:
kfp-api creates an extra Service object called `ml-pipeline` to be compatible with upstream Kubeflow. On Juju versions < 2.9, this Service object used a selector of `juju-app: kfp-api`. However, starting with Juju 2.9, that selector needs to be `app.kubernetes.io/name: kfp-api` instead.

Fixes #6

### DIFF
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -258,7 +258,9 @@ class Operator(CharmBase):
                         {
                             "name": "ml-pipeline",
                             "spec": {
-                                "selector": {"juju-app": self.model.app.name},
+                                "selector": {
+                                    "app.kubernetes.io/name": self.model.app.name
+                                },
                                 "ports": [
                                     {
                                         "name": "grpc",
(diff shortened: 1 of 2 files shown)


joedborg (approved):
LGTM

PR #20, opened 2021-09-06
Add license file

DomFleischmann:
Addresses #18


evilnick:
Again, although github says it isn't necessary, the Apache license recommends a boilerplate. It seems excessive to add this to each file but as a separate COPYRIGHT file that would be okay.

Are we also going to transition the 'master' branch to 'main' here too?

PR #24, opened 2021-11-26
Update kfp-api relation/status handling, tests

ca-scribner:


charms/kfp-api/src/charm.py
```diff
@@ -13,141 +15,55 @@
...
-            )
-            self.framework.observe(
-                self.on[relation].relation_changed,
-                self.send_info,
-            )
-
-    def send_info(self, event):
-        if self.interfaces["kfp-api"]:
-            self.interfaces["kfp-api"].send_data(
+    def send_info(self, interfaces):
```
DomFleischmann, line 42: This method could be made private. I that in most of the charms

charms/kfp-api/src/charm.py
```diff
@@ -286,98 +202,180 @@ def set_pod_spec(self, event):
...
+            # relations is a dict of {(ops.model.Relation, ops.model.Application): data}
+            unpacked_relation_data = relations.get_data()
+        except ValidationError as val_error:
+            # Validation in .get_data() ensures if data is populated, it matches the schema and is
+            # not incomplete
+            self.log.error(val_error)
+            raise CheckFailed(
+                f"Found incomplete/incorrect relation data for {relation_name}.  See logs",
+                BlockedStatus
+            )
```
DomFleischmann, line 352: This seems like something where I'm comfortable with going into an error status instead of blocked, if that happens something clearly went wrong and an error status seems a lot more appropiate than blocked.

`BlockedStatus` is for when something needs human intervention, but not for any kind of error, I would prefer if the `ValidationError` is not catched and the charm fails here.

charms/kfp-api/src/charm.py
```diff
@@ -286,98 +202,180 @@ def set_pod_spec(self, event):
...
+        if len(unpacked_relation_data) == 0:
+            raise CheckFailed(f"Waiting for {relation_name} relation data", WaitingStatus)
+
+        # Unpack data (we care only about the first element)
+        data_dict = list(unpacked_relation_data.values())[0]
+
+        # Catch if empty data dict is received (JSONSchema ValidationError above does not raise
+        # when this happens)
+        if len(data_dict) == 0:
+            raise CheckFailed(f"Found incomplete/incorrect relation data for {relation_name}.", BlockedStatus)
```
DomFleischmann, line 364: This seems more like a bug in SDI than something that we should be handling as an edge case in our charms.

charms/kfp-api/src/charm.py
```diff
@@ -286,98 +202,180 @@ def set_pod_spec(self, event):
...
+        # If nothing is related to this relation, return a default value or raise an error
+        if relation_name not in interfaces or interfaces[relation_name] is None:
+            if default_return is not None:
+                return default_return
+            else:
+                raise CheckFailed(f"Missing required relation for {relation_name}", BlockedStatus)
+
+        relations = interfaces[relation_name]
+        if not isinstance(relations, SerializedDataInterface):
+            raise CheckFailed(f"Unexpected error with {relation_name} relation data - data not as expected", BlockedStatus)
```
DomFleischmann, line 339: This again seems like something that I prefer erroring out instead of us handling it.

DomFleischmann (commented):
I like how the unit tests look.

My only issue is with some parts of `validate_sdi_interface`, while I understand the intention of trying to cover all the different cases where a charm could fail i don't think it is the right approach for us due to the following reasons:
1. Blocked and Error states both require manual intervention, therefore we are in both cases in very similar situations. But the key difference between them is that blocked happens when the charm works how it should, but some additional step from the user is required. Error is when something is not working how it should. An `ValidationError` or `interfaces[relation_name]` not being a `SerializedDataInterface` instance are clearly occasions where the charm is not working as intended and therefore it should error out.
2. By hiding these exceptions behind `CheckFailed` exceptions we lose the stack traces and actually get less information than if we had just allowed them to be raised normally.
3. If SDI is not resilient enough (which is probably the case) we should not cover that up in our charm code but actually fix SDI.

charms/kfp-api/src/charm.py
ca-scribner: I was about to completely agree with this suggestion, but then thought about something like Istio with optional relations and now I'm not sure.  

Agreed we're really catching an **error** here.  If we let this raise to Juju, no further hooks fire and the charm is in stasis.  That would be fine for a required relation like here, but not an optional relation like in Istio.  I lean toward using the same pattern for both required and optional relations.  What do you think?  It is already hard to keep our many charms consistent - the fewer patterns the better imo.  Functionally, the effect of Blocked vs Error on this relation feels equivalent (they both show up loudly on `juju status` and both convey "we're broken!").  

[Juju docs](https://juju.is/docs/olm/status-values) are a bit ambiguous about the best way to do things.  My interpretation is that charms handle things they expect might happen in `Blocked`, and `Error` is for the rest.  A related charm giving me garbage data feels like something I should expect possible and handle.

charms/kfp-api/src/charm.py
ca-scribner: Agreed.  I don't want to take it out because I think it can happen, but I'll also file it on SDI to make sure it is prevented

charms/kfp-api/src/charm.py
ca-scribner: This should be resolved the same as the above conversation.  In this case, I am not sure this can happen if SDI works correctly anyway.

charms/kfp-api/src/charm.py
ca-scribner: Replying to the general review comments here so I don't splinter the conversation

Re: Blocked meaning the charm worked the way it should and Error meaning the charm didn't, I agree completely.  But in the case here I think **this** charm is still working the way it should.  We can have a perfect charm, but if some other charm sends us bad data and we notice it that feels like we're blocked (given bad inputs/config) than errored

Re hiding the exceptions behind CheckFailed, agreed that could happen but I don't think it happens here.  By catching `ValidationError` we're doing a value add rather than hiding anything.  We [log the error](https://github.com/canonical/kfp-operators/pull/24/files#diff-5e034222c66adb8d88c366d84692add5e82b27063ffa982293ac8328380339d3R348) so we haven't lost anything, and we [add context saying exactly what relation broke](https://github.com/canonical/kfp-operators/pull/24/files#diff-5e034222c66adb8d88c366d84692add5e82b27063ffa982293ac8328380339d3R350).  I think by handling the error we come out ahead.  We could still catch, log context, and reraise the `ValidationError` instead of `CheckFailed`, but I think we get more this way.

charms/kfp-api/src/charm.py
```diff
@@ -286,98 +204,189 @@ def set_pod_spec(self, event):
...
+
+        # Unpack data (we care only about the first element)
+        data_dict = list(unpacked_relation_data.values())[0]
+
+        # Catch if empty data dict is received (JSONSchema ValidationError above does not raise
+        # when this happens)
+        if len(data_dict) == 0:
+            raise CheckFailedError(
+                f"Found incomplete/incorrect relation data for {relation_name}.",
+                BlockedStatus,
```
DomFleischmann, line 374: Same as the other [review](https://github.com/canonical/kfp-operators/pull/27#discussion_r759448470)

DomFleischmann (approved):
Just the same comment as in the others

PR #40, opened 2021-12-07
Test publish.yaml manual process

ca-scribner:


.github/workflows/publish-bundle.sh
```diff
@@ -0,0 +1,18 @@
...
+BUNDLE_FILE=$(cat publishlog_charmcraft_pack.log | cut -d "'" -f 2)
+echo Found BUNDLE_FILE=$BUNDLE_FILE
+
+charmcraft upload $BUNDLE_FILE | tee publishlog_charmcraft_upload.log
+
+# Parse revision number from upload output
+REVISION=$(cat publishlog_charmcraft_upload.log | cut -d " " -f 2)
+echo Found REVISION=$REVISION
+
+charmcraft release $BUNDLE_NAME --revision $REVISION --channel=edge
```
DnPlas, line 18: I'm not entirely sure if there's a distinction between `latest/edge` and just `edge`. If they are the same, ignore this comment.

.github/workflows/publish.yaml
```diff
@@ -4,29 +4,50 @@ on:
...
-          - kfp-api
-          - kfp-persistence
-          - kfp-profile-controller
-          - kfp-schedwf
-          - kfp-ui
-          - kfp-viewer
-          - kfp-viz
+      matrix: ${{ fromJson(needs.get-charm-matrix.outputs.charm_paths_json) }}
     steps:
       - uses: actions/checkout@v2
```
DnPlas, line 35: The conditional statement `if push || branch_name == branch/*` applies here

bundle.yaml
```diff
@@ -1,14 +1,15 @@
 bundle: kubernetes
+name: kubeflow-pipelines
 applications:
-  argo-controller:        { charm: cs:~argo-charmers/argo-controller, scale: 1 }
-  minio:                  { charm: cs:~minio-charmers/minio-13,       scale: 1 }
-  kfp-api:                { charm: cs:kfp-api,                        scale: 1 }
+  argo-controller:        { charm: argo-controller, channel: latest/edge, scale: 1 }
```
DnPlas, line 4: Missing `ch`?

DnPlas (commented):
Looks good to me so far, just a couple of comments. Thanks @ca-scribner 

Future work: Every time you make a change in a single charm, all other charms will be published again despite being the same. Though this is okay for our initial publish job, we can make it more selective and smarter.

.github/workflows/publish-bundle.sh
ca-scribner: I'm not sure either.  I worried there might be, so I put it in...

PR #47, opened 2021-12-14
Remove unneeded deployment from metadata.yaml

ca-scribner:
Remove deployment field from metadata.yaml in kfp-api, kfp-profile-controller, kfp-ui, and kfp-viz


DomFleischmann (approved):
LGTM

PR #59, opened 2022-02-09
fix: ensure minio creds in namespace are updated on change

### LINKED ISSUE #56: `mlpipeline-minio-artifact` secret out of sync with actual minio credentials
As discussed in #53, the `mlpipeline-minio-artifact` secret which is managed by a `compositecontroller` metacontroller + the `sync.py` script deployed by the `kfp-profile-controller` charm as the `kfp-profile-controller` pod.  If the `sync.py`/`kfp-profile-controller` workload is updated to have a new value for this secret, the `mlpipeline-minio-artifact` secret **is not automatically updated**.  This results in pipeline steps failing as they use this secret for their minio credentials.

ca-scribner:
Fixes an issue where the metacontroller defined by the kfp-profile-controller does not update the object storage credentials (stored in the `mlpipeline-minio-artifact` secret in each user's namespace), even though the kfp-profile-controller itself is informed of the new creds.  This was caused by using the `OnDelete` update mode in the metacontroller, which will only update the object if it is fully deleted (and not if it has just changed).  This is now set to `InPlace`, which always updates the object if the desired state differs from the current state (similar to if you did a `kubectl apply`).

Closes #56

### DIFF
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -136,7 +136,8 @@ class KfpProfileControllerOperator(CharmBase):
                                         {
                                             "apiVersion": "v1",
                                             "resource": "secrets",
-                                            "updateStrategy": {"method": "OnDelete"},
+                                            # TODO: Push this change upstream
+                                            "updateStrategy": {"method": "InPlace"},
                                         },
                                         {
                                             "apiVersion": "v1",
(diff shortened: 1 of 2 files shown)


DomFleischmann (commented):
This LGTM to me, the test that failed seemed to be something flacky as it was minio not being able to pull its image. Lets re run them and see if they pass

PR #60, opened 2022-02-10
add min-juju-version to metadata

VariableDeclared:
add min-juju-version to prevent PVCs being created

### DIFF
--- a/charms/kfp-api/metadata.yaml
+++ b/charms/kfp-api/metadata.yaml
@@ -4,6 +4,7 @@ description: |
   Machine learning (ML) toolkit that is dedicated to making deployments
   of ML workflows on Kubernetes simple, portable, and scalable.
 series: [kubernetes]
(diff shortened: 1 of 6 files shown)


charms/kfp-api/metadata.yaml
```diff
@@ -4,6 +4,7 @@ description: |
   Machine learning (ML) toolkit that is dedicated to making deployments
   of ML workflows on Kubernetes simple, portable, and scalable.
 series: [kubernetes]
+min-juju-version: "2.8.0"
```
DomFleischmann, line 7: Why are these ones `2.8.0` and in other PRs its `2.9.0`?

VariableDeclared: it's an escapee. let me fix that

DomFleischmann (approved):

PR #63, opened 2022-02-17
Updated access and secret key handling [KF-208] and added more integration tests [KF-212]

natalian98:
1. Modified access and secret key handling to pass them as secrets instead of environment variables, therefore not exposing them in plain text [KF-208]
2. Expanded the integration tests for kfp-profile-controller [KF-212]
- Deploy kubeflow-profiles and metacontroller-operator required for CRDs creation
- Test if the charms are deployed and active
- Create a test profile and corresponding namespace with label pipelines.kubeflow.org/enabled=true
- Ensure that the default-editor service account was created
- Check whether the secret exists and contains the expected key
- Ensure that the pods are ready
3. Modified charm.py due to black lint formatting
4. Added juju agent version in bootstrap options (NOTE: to be removed once juju is fixed, see [KF-231])

[KF-208]: https://warthogs.atlassian.net/browse/KF-208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KF-212]: https://warthogs.atlassian.net/browse/KF-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KF-231]: https://warthogs.atlassian.net/browse/KF-231


charms/kfp-profile-controller/src/charm.py
```diff
@@ -70,8 +72,7 @@ def _set_pod_spec(self, event):
         self.model.unit.status = MaintenanceStatus("Setting pod spec")
 
         deployment_env = {
-            "MINIO_ACCESS_KEY": os["access-key"],
-            "MINIO_SECRET_KEY": os["secret-key"],
+            "minio-secret-kfp": {"secret": {"name": "minio-secret-kfp"}},
```
ca-scribner, line 75: I read this as minio's secret for kfp, not kfp's secret for minio.  Maybe flip it around?  This could also get deployed more than once in the same namespace in theory, so maybe better practice would be to do `CHARM_NAME-minio-secret` or similar?

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -164,7 +167,11 @@ def test_install_with_all_inputs(harness, oci_resource_data):
     harness.add_relation_unit(os_rel_id, "storage-provider/0")
     harness.update_relation_data(os_rel_id, "storage-provider", os_data)
 
-    harness.begin_with_initial_hooks()
+    pod_spec = harness.get_pod_spec()
```
ca-scribner, line 170: Since you create the `pod_spec` here, we can reuse `pod_spec` in the assertion a few lines after this rather than invoke it again.

charms/kfp-profile-controller/src/charm.py
ca-scribner: (eg: `"{self.model.app.name}-minio-credentials"`, or similar)

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -39,13 +39,22 @@ def emit_settings_to_logs(settings):
...
 
 
-def get_settings_from_env(controller_port=None,
-                          visualization_server_image=None, frontend_image=None,
-                          visualization_server_tag=None, frontend_tag=None,
-                          disable_istio_sidecar=None, minio_access_key=None, minio_secret_key=None,
-                          minio_host=None, minio_port=None, minio_namespace=None,
-                          kfp_default_pipeline_root=None, metadata_grpc_service_host=None,
-                          metadata_grpc_service_port=None):
+def get_settings_from_env(
```
ca-scribner, line 42: Are the changes in this file just to satisfy the linting/formatting, or are there substanttitive changes here too?  This file comes from upstream, so to make our lives easier when reconciling changes with upstream in the future we should avoid formatting changes here.  If this was to satisfy out code checks, we should add an exemption for this file

charms/kfp-profile-controller/requirements.txt
```diff
@@ -1,3 +1,5 @@
 ops==1.2.0
 oci-image==1.0.0
 serialized-data-interface<0.4   
+lightkube<0.9
+tenacity<8.1
```
ca-scribner, line 5: Are these used in the charm, or just in the tests?  It looks like just in the tests.  If so, they can go [here](https://github.com/natalian98/kfp-operators/blob/c40f729ed26d509d25c50e8e048d16f96184e884/charms/kfp-profile-controller/tox.ini#L70) instead, as putting them in this `requirements.txt` will ship them with the deployed charm

charms/kfp-profile-controller/tests/integration/test_charm.py
```diff
@@ -35,19 +44,181 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+
+
+async def test_status(ops_test: OpsTest):
+    charms = ("kubeflow-profiles", "metacontroller-operator", "kfp-profile-controller")
+    for charm in charms:
+        assert ops_test.model.applications[charm].units[0].workload_status == "active"
+
+
+async def test_profile_creation(lightkube_client, profile):
+    # Test whether a namespace was created for this profile
```
ca-scribner, line 68: This makes me think we're testing whether the profile was created, but I think the intent is to test whether a `Namespace`, etc, was created for the given `Profile`?  Maybe reword (and make the headline comment a docstring instead of a comment).  Please reword as you see fit, just a suggested wording
```suggestion
async def test_profile_resource_creation(lightkube_client, profile):
    """Test whether resources expected for a profile were created."""
```

charms/kfp-profile-controller/tests/integration/test_charm.py
```diff
@@ -35,19 +44,181 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    profile_name = profile
+    validate_profile_resources(lightkube_client, profile_name)
```
ca-scribner, line 70: Small nit similar to that above.  The naming here makes me think you're validating something about the `Profile`, not about the resources you're creating for the profile.  If you can think of a better name that's great, otherwise leave as is

charms/kfp-profile-controller/tests/integration/test_charm.py
```diff
@@ -35,19 +44,181 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+def _validate_if_exists(if_exists):
+    if if_exists in ALLOWED_IF_EXISTS:
+        return if_exists
+    else:
+        raise ValueError(
+            f"Invalid value for if_exists '{if_exists}'.  Must be one of {ALLOWED_IF_EXISTS}"
+        )
+
+
+def create_all_from_yaml(
```
ca-scribner, line 115: I know who to blame for this function (I believe you pulled it from a past charm of mine :D ), but now that `lightkube` provides an [`apply` function](https://lightkube.readthedocs.io/en/latest/client/) I think this could just use something native from `lightkube`.  Give it a thought and see if you can simplify this

charms/kfp-profile-controller/tests/integration/test_charm.py
ca-scribner: oh wait I just misunderstood the test.  You are actually testing whether the `Profile`'s stuff was created here, not whether the things you create (secrets, etc) are present

ca-scribner: This is more subjective, but I think this testing functionality would make more sense inside the 'profile' fixture. My gut says tests here test things your charm is doing, and the fixtures are for providing stuff that comes from elsewhere, so seeing a fixture fail would indicate something broke in my dependencies. Having said that though, there's some flexibility here.

I'm pretty sure pytest-operator runs tests linearly so you can always guarantee this test runs before anything below it, but putting the test/raise if it breaks functionality into the fixture itself might make it more clear

.github/workflows/integrate.yaml
```diff
@@ -62,6 +62,7 @@ jobs:
         with:
           provider: microk8s
           channel: 1.21/stable
+          bootstrap-options: --agent-version="2.9.22"
```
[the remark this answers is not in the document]
ca-scribner: yeah that would be perfect

charms/kfp-profile-controller/files/upstream/sync.py
natalian98: It was to satisfy black lint requirements. I will revert the changes and add the upstream file to exceptions.

charms/kfp-profile-controller/requirements.txt
natalian98: Right! Sorry for my ignorance on tox. I will move those here.

charms/kfp-profile-controller/tests/integration/test_charm.py
natalian98: I'm not sure I can use the `apply` function. According to lightkube documentation, `server-side apply` requires k8s v1.22 while we're using 1.21. Please correct me if I'm wrong.

ca-scribner: Based on discussion in our daily, I think this is unblocked now and it `apply` should work

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -486,4 +486,4 @@ def do_POST(self):
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
ca-scribner, line 489: I think this is missing the newline at the end that was there before?
```suggestion
    main()

```

charms/kfp-profile-controller/tests/integration/test_charm.py
```diff
@@ -35,19 +43,125 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
-#  * check whether secrets and other things get updated based on the config sent to sync.py/kfppc
-#  * could create namespaces with the right pipelines label to trigger the sync, but also need to
-#    create a dummy poddefaults CRD too
+    await ops_test.model.wait_for_idle(status="active", timeout=60 * 10)
+
+
+async def test_status(ops_test: OpsTest):
+    charms = ("kubeflow-profiles", "metacontroller-operator", "kfp-profile-controller")
+    for charm in charms:
+        assert ops_test.model.applications[charm].units[0].workload_status == "active"
```
ca-scribner, line 63: I think this is redundant with line 57

PR #64, opened 2022-02-23
encode minio access and secret key

agathanatasha:
encode minio access and secret key in base64 and store in kubernetes secret

### DIFF
--- a/charms/kfp-ui/src/charm.py
+++ b/charms/kfp-ui/src/charm.py
@@ -9,6 +9,7 @@ https://github.com/canonical/kfp-operators
 
 import json
 import logging
+from base64 import b64encode
 
 from jsonschema import ValidationError
 from oci_image import OCIImageResource, OCIImageResourceError
@@ -88,11 +89,11 @@ class KfpUiOperator(CharmBase):
             "KUBEFLOW_USERID_PREFIX": "",
             "METADATA_ENVOY_SERVICE_SERVICE_HOST": "localhost",
             "METADATA_ENVOY_SERVICE_SERVICE_PORT": "9090",
-            "MINIO_ACCESS_KEY": os["access-key"],
+            "MINIO_ACCESS_KEY": b64encode(os["access-key"].encode("utf-8")).decode("utf-8"),
             "MINIO_HOST": os["service"],
             "MINIO_NAMESPACE": os["namespace"],
             "MINIO_PORT": os["port"],
-            "MINIO_SECRET_KEY": os["secret-key"],
+            "MINIO_SECRET_KEY": b64encode(os["secret-key"].encode("utf-8")).decode("utf-8"),
             "MINIO_SSL": os["secure"],
             "ML_PIPELINE_SERVICE_HOST": kfp_api["service-name"],
             "ML_PIPELINE_SERVICE_PORT": kfp_api["service-port"],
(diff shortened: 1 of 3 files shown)


charms/kfp-ui/src/charm.py
```diff
@@ -88,11 +89,10 @@ def _main(self, event):
             "KUBEFLOW_USERID_PREFIX": "",
             "METADATA_ENVOY_SERVICE_SERVICE_HOST": "localhost",
             "METADATA_ENVOY_SERVICE_SERVICE_PORT": "9090",
-            "MINIO_ACCESS_KEY": os["access-key"],
+            "minio-secret": {"secret": {"name": "minio-secret"}},
```
ca-scribner, line 92: Re the name of the secret, with this name the secret will conflict with the secret our minio charm creates for itself.  This should be named something like `CHARM_NAME-credentials`

Also, by statically defining it (eg: `kfp-ui-credentials`) rather than dynamically (eg: `"{CHARM_NAME}-credentials"`) we also have a problem if we deploy the same charm twice in one model.  That probably isn't an issue here, but it could be.

charms/kfp-ui/tests/unit/test_operator.py
```diff
@@ -297,6 +298,70 @@ def test_install_with_all_inputs(harness, oci_resource_data):
...
+
+    # object storage relation
+    os_data = {
+        "_supported_versions": "- v1",
+        "data": yaml.dump(
+            {
+                "access-key": "access-key",
+                "namespace": "namespace",
+                "port": 1234,
+                "secret-key": "secret-key",
```
ca-scribner, line 324: really nitpicky, but I always worry in tests about setting keys and values to the same thing (eg: `{"access-key": "access-key"}`), it can accidentally hide bugs where you use the wrong one by mistake (eg: you meant to use `dict.values` but instead used `dict.keys`)

ca-scribner: (also, the irony is not lost on me that I bet you followed the pattern from the other tests, and `git blame` points to me for that mistake :) )

charms/kfp-ui/tests/unit/test_operator.py
```diff
@@ -297,6 +298,70 @@ def test_install_with_all_inputs(harness, oci_resource_data):
...
+    harness.update_relation_data(
+        kfpui_rel_id, f"{kfpui_relation_name}-subscriber", relation_version_data
+    )
+
+    harness.begin_with_initial_hooks()
+
+    pod_spec, _ = harness.get_pod_spec()
+    minio_secrets = pod_spec["kubernetesResources"]["secrets"][0]["data"]
+    assert b64decode(minio_secrets["MINIO_ACCESS_KEY"]).decode("utf-8") == "access-key"
+    assert b64decode(minio_secrets["MINIO_SECRET_KEY"]).decode("utf-8") == "secret-key"
```
ca-scribner, line 362: I'm not sure which is better, but is there a reason why we wouldn't put this test in the above `test_install_with_all_inputs` instead?  that is sort of an "end to end unit test" that does almost everything you need

ca-scribner (approved):

charms/kfp-ui/tests/unit/test_operator.py
ca-scribner: yeah it really is a style choice rather than there being right/wrong.  Where there's so much duplication I lean toward merging but ...

PR #68, opened 2022-03-25
Modified grafana template, added kfp-api test to CI

natalian98:
- Modified grafana template to meet the charm requirements (a valid json instead of empty file)
- Added kfp-api integration test to CI


charms/kfp-api/src/charm.py
```diff
@@ -40,18 +42,19 @@ def __init__(self, *args):
...
-                    "scrape_interval": self.config["metrics-scrape-interval"],
-                    "metrics_path": self.config["metrics-api"],
+                    "metrics_path": METRICS_PATH,
                     "static_configs": [{"targets": ["*:{}".format(self.config["http-port"])]}],
                 }
             ],
         )
 
-        self.dashboard_provider = GrafanaDashboardProvider(self)
+        self.dashboard_provider = GrafanaDashboardProvider(
```
agathanatasha, line 54: don't need to change this line now that we are using the default name for the relation

charms/kfp-api/lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -21,8 +21,9 @@
 use case of a charm which bundles dashboards and provides a
 `provides: grafana-dashboard` interface,
 
-    grafana-dashboard:
-      interface: grafana_dashboard
+    requires:
```
agathanatasha, line 24: should we update this and the prometheus_scrape lib files in the other repos? I haven't been doing this

natalian98: That was recommended by Observability Team. [Updated version](https://github.com/canonical/grafana-k8s-operator/blob/main/lib/charms/grafana_k8s/v0/grafana_dashboard.py) handles some exceptions I had in the previous one.

charms/kfp-api/src/charm.py
```diff
@@ -40,18 +42,19 @@ def __init__(self, *args):
...
         self.prometheus_provider = MetricsEndpointProvider(
             charm=self,
-            relation_name="monitoring",
+            relation_name="metrics-endpoint",
             jobs=[
                 {
-                    "job_name": "ml_pipeline",
-                    "scrape_interval": self.config["metrics-scrape-interval"],
-                    "metrics_path": self.config["metrics-api"],
+                    "metrics_path": METRICS_PATH,
```
ca-scribner, line 48: Doesn't hurt to have this, but also this is the default anyway.  This can be removed

charms/kfp-api/src/charm.py
ca-scribner: Yeah I think what @agathanatasha is suggesting is to do this:
```suggestion
        self.dashboard_provider = GrafanaDashboardProvider(self)
```

charms/kfp-api/tests/integration/test_charm.py
```diff
@@ -61,26 +63,52 @@ async def test_prometheus_grafana_integration(ops_test: OpsTest):
...
+    scrape_config = {"scrape_interval": "5s"}
 
     await ops_test.model.deploy(prometheus, channel="latest/beta")
     await ops_test.model.deploy(grafana, channel="latest/beta")
     await ops_test.model.add_relation(prometheus, grafana)
     await ops_test.model.add_relation(APP_NAME, grafana)
-    await ops_test.model.add_relation(prometheus, APP_NAME)
+    await ops_test.model.deploy(
+        prometheus_scrape_charm,
+        channel="latest/beta",
```
ca-scribner, line 75: Why do we use latest/beta here?  This sounds like maybe we need a feature that'll be promoted later?  If yes, add a comment here and link an issue in the repo saying when we can remove this and pull from latest/stable again.  

In that issue, add a note to do the same for the testing instructions readme, too :)

ca-scribner (approved):
lgtm, but there's a few things that could be changed.  see comments

charms/kfp-api/src/charm.py
natalian98: Can it be kept here since other charms also have this specified, even though it's default?

charms/kfp-api/tests/integration/test_charm.py
natalian98: This charm is not available in stable. @mmanciop please correct me if I'm wrong, but I think latest/beta is currently the only channel it can be fetched from?

ca-scribner: Let's not hold merge up on this.  We can keep this and merge, but just add a comment/gh issue that reminds us to update when there's a stable version

natalian98: Sure, I created a gh issue for this. Thanks

PR #69, opened 2022-04-01
Observe leader_settings_changed in kfp-ui and kfp-profile-controller

### LINKED ISSUE #52: Units get stuck waiting for leadership after a ConfigChanged event
On a fresh CKF deployment, when a ConfigChanged event is triggered, `kfp-*` units can get stuck in `WaitingStatus("Waiting for leadership")`.

This happened to me after changing `minio` access and secret keys. Since pipelines relate to MinIO, modifying keys will trigger ConfigChanged events in `kfp-ui` and `kfp-profile-controller`.

#### Steps to reproduce
1. Deploy Charmed Kubeflow v1.4 in Microk8s 1.21, wait for all charms in the bundle to be active
2. Change MinIO keys: `juju config access-key soMeS3cret secret-key 0th3rKey`
3. New `kfp-ui` and `kfp-profile-controller` units are created (existing ones are removed)
4. `juju status` shows `kfp-ui` and `kfp-profile-controller` in a waiting status with message "Waiting for leadership". This doesn't change with time.

natalian98:
closes #52

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build/
 __pycache__/
 .tox
 .coverage
+.idea
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -56,7 +56,7 @@ class KfpProfileControllerOperator(CharmBase):
         self.framework.observe(self.on.upgrade_charm, self._set_pod_spec)
         self.framework.observe(self.on.config_changed, self._set_pod_spec)
         self.framework.observe(self.on["object-storage"].relation_changed, self._set_pod_spec)
-        self.framework.observe(self.on.leader_settings_changed, self._set_pod_spec)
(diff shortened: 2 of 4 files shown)


agathanatasha:
From what I understand, leader-settings-changed is only emitted for non-leader units and leader-elected is only emitted for leader units. If the issue is about unit stuck in waiting status, shouldn't we observe leader-elected event instead?

natalian98:
> From what I understand, leader-settings-changed is only emitted for non-leader units and leader-elected is only emitted for leader units. If the issue is about unit stuck in waiting status, shouldn't we observe leader-elected event instead?

Hi @agathanatasha, thanks for checking this PR, much appreciated! I think you are right, documentation says the same. But both `leader_settings_changed` and `leader_elected` seem to work likewise: I compared the logs and they both emit the same hooks.
@dnplas @ca-scribner what do you think?

ca-scribner (commented):
lgtm

PR #79, opened 2022-07-18
bump kf-schedwf version to 2.0.0-alpha.3 

ca-scribner:
Note this is merging into a feature branch, not main, as it should not be released without the rest of kfp v2.0.0

Also:
* adds access to scheduledworkflow finalizers
* updates the scheduledworkflow CRD


DomFleischmann (changes requested):
You seem to have a lint error in charm.py

DomFleischmann (approved):

PR #82, opened 2022-07-18
update config options for kfp-profile-controller per kfp v2.0.0-alpha.3

ca-scribner:
Note this is merging into a feature branch, not main, as it should not be released without the rest of kfp v2.0.0

This updates the profile controller to kfp v2.0.0-alpha.3.  No workload needs to be updated as this component has no binary workload


DomFleischmann (commented):
There is the hardcoded `KFP_IMAGES_VERSION` variable in `charm.py` which still points to `"1.7.0-rc.3"` should that be updated?

Also can you pull the latest observability lib?

PR #85, opened 2022-07-26
feat: convert bundle tests to pytest-operator

ca-scribner:
Refactors the bundle tests to use pytest-operator for deployment, updating the previous method of "manually" deploying the bundle from `integrate.yaml`


tests/integration/helpers/localize_bundle.py
```diff
@@ -0,0 +1,98 @@
...
+def get_charm_name(metadata_file: Union[Path, str]) -> str:
+    metadata = yaml.safe_load(Path(metadata_file).read_text())
+    return metadata["name"]
+
+
+def get_charm_file(charm_dir: Path) -> Path:
+    """Returns the path to the .charm file representing the charm in the given directory
+
+    TODO: This just assumes the suffix on the file name will be "ubuntu-20.04-amd64".
+          Fix this in future
```
DomFleischmann, line 22: Might be a bit overkill and not a long term fix, but you could check the bases in the `charmcraft.yaml` of each charm.

tests/integration/test_bundle.py
```diff
@@ -0,0 +1,75 @@
...
+    # TODO This should be easier... but it isn't.  Can either use the `juju bundle` integration, or
+    #      shell out to juju (like k8s charms do).  For now, I'm doing the latter
+    #      Observability has a function in their CI that is similar.  Use that?
+    bundle_dst_file = write_bundle_file(bundle, ops_test.tmp_path / "bundles" / "bundle.yaml")
+
+    model = ops_test.model_full_name
+    # TODO: Figure out when we need --trust rather than blanket-applying it
+    cmd = f"juju deploy -m {model} --trust {bundle_dst_file}"
+    log.info(f"Deploying bundle to {model} using cmd '{cmd}'")
+    rc, stdout, stderr = await ops_test.run(*shlex.split(cmd))
```
DomFleischmann, line 56: All of this seems like something we should add to pytest-operator and standardize somehow.

DomFleischmann (approved):
Some comments about some of the helpers, but this looks very good.

tests/integration/test_bundle.py
ca-scribner: 100% agree.  pytest-operator has some similar features, but they're a bit different.  They don't do this "convert a publishable bundle into partly-local", so I think this would help.  I'm thinking we gain a little experience then push the final product to pytest-operator?

tests/integration/helpers/localize_bundle.py
ca-scribner: Leaving as is for now, but we should do this in future.  Probably this can be pulled from `charmcraft.yaml`, but `arch` is not obviously a property in there.  There's a question outstanding to the starcraft team

PR #86, opened 2022-07-27
fix: kfp-api observability integration tests

ca-scribner:
Updates test to explicitly call out relations.  Refactors tests for consistency with others.

### DIFF
--- a/charms/kfp-api/tests/integration/test_charm.py
+++ b/charms/kfp-api/tests/integration/test_charm.py
@@ -83,7 +83,7 @@ async def test_prometheus_grafana_integration(ops_test: OpsTest):
         f"{prometheus}:metrics-endpoint", f"{prometheus_scrape}:metrics-endpoint"
     )
 
-    await ops_test.model.wait_for_idle(status="active", timeout=60 * 10)
+    await ops_test.model.wait_for_idle(status="active", timeout=60 * 20)
 
     status = await ops_test.model.get_status()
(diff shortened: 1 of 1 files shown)


ca-scribner:
Note that this is against kf-16-release, not main, so publishing jobs will fail

charms/kfp-api/tests/integration/test_charm.py
```diff
@@ -61,23 +61,29 @@ async def test_build_and_deploy(ops_test: OpsTest):
 
 async def test_prometheus_grafana_integration(ops_test: OpsTest):
     """Deploy prometheus, grafana and required relations, then test the metrics."""
+    application_name = APP_NAME
```
natalian98, line 64: Do we need `application_name` if we can just use `APP_NAME`? Later in this test, we use `APP_NAME` anyway for testing the prometheus deployment

ca-scribner: ahh I missed that it was used more.  I localised it because I copied/pasted some code from elsewhere that took `application_name` as an arg, but I'll change it here.

natalian98:
lgtm, but since it's for 1.6 release, maybe we could test on microk8s 1.22? What do you think?

ca-scribner:
Totally agree that testing on microk8s 1.22 is a good idea, but I'll do it separately to keep cause and effect of any breaks clear if that's ok

PR #92, opened 2022-08-04
update libs

agathanatasha:


ca-scribner (approved):
Tests are failing in kfp-api because of [this bug](https://github.com/juju/python-libjuju/pull/696), which affects the current `prometheus-k8s` `latest/edge`.  This can be fixed by pinning to juju in github, waiting till the next python-libjuju release, or temporarily commenting out the prometheus tests

PR #93, opened 2022-08-05
feat: add manually-triggerable publish action

ca-scribner:
This makes the publish action clickable manually in the Actions tab of github, allowing users to manually publish a charm from gh:main->ch:latest/edge


natalian98:
Looks great @ca-scribner.
I guess the tests are failing because of recent split of observability libs.
Could you fetch [this](https://github.com/canonical/notebook-operators/tree/master/charms/jupyter-controller/lib/charms/observability_libs/v0) library and update others if necessary? Or do you think we should open a separate PR for this?

ca-scribner:
@natalian98 turns out once I tested in a temporary repo this approach doesn't work... I'm refactoring now

I think the libs should be updated separately.  No need to combine them

ca-scribner:
I think this should work better.  I've tested it [in this repo](https://github.com/ca-scribner/test-multi-charm), with passing CI for both [on push](https://github.com/ca-scribner/test-multi-charm/actions/runs/2820674909) and [on manual trigger](https://github.com/ca-scribner/test-multi-charm/actions/runs/2820818530)

.github/workflows/publish.yaml
```diff
@@ -1,41 +1,90 @@
...
       matrix:
-        charm:
-          - kfp-api
-          - kfp-persistence
-          - kfp-profile-controller
-          - kfp-schedwf
-          - kfp-ui
-          - kfp-viewer
-          - kfp-viz
+        charm-path: ${{ fromJson(needs.get-charm-paths.outputs.charm_paths_list) }}
```
natalian98, line 41: That's great, thanks for implementing this!
Just to keep consistency across all workflows, could you change the matrix of hardcoded charms in `integration.yaml` to use the `get-charm-paths` action too?

.github/workflows/get-charm-paths.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash -x
```
agathanatasha, line 1: I like this for figuring the charm paths, but I think it should live in some shared repos. Maybe put it in charming actions?

.github/workflows/publish.yaml
natalian98: Giving this a second thought, we should handle that part separately. For example, kfp has integration tests only for 2 charms, so it doesn't make sense to use script in lint and unit but not for integration within the same workflow.

.github/workflows/get-charm-paths.sh
ca-scribner: 100% agreed.  Only question to me is schedule - do we have time to make it into an action, or should we push this first and then later replace with an action?  I'll poke @DomFleischmann

ca-scribner: Discussed with @DomFleischmann and we will push the script for now, but make this into a real action in the near term

natalian98 (approved):

PR #95, opened 2022-08-10
Observe leader_elected event

natalian98:
When we use `juju upgrade-charm`, a new non-leader unit gets created, then the old still-leader one gets removed, resulting in the new unit being elected as the leader.
When the `leader_elected` event is not observed, the charm is not aware of having a leader now, so it remains stuck with message 'Waiting for leadership'.

Note: I reverted unit test changes as this will be handled separately once agreed on [reusable tests](https://github.com/canonical/charmed-kubeflow-chisme/pull/23).


DnPlas:
LGTM, thanks @natalian98

DnPlas (approved):

PR #96, opened 2022-08-12
cannon/change-eae484fd

ca-scribner:
Changes applied by commit-cannon:
  * Replaced file `.github/workflows/get-charm-paths.sh`
  * Replaced file `.github/workflows/publish.yaml`


ca-scribner:
this was a test

PR #111, opened 2022-11-22
feat: bump track 2.0 to kfp released for KF 1.6.1, fix CI

ca-scribner:
This bumps kfp to the version released with the 1.6.1 Kubeflow release. It also fixes CI to address backwards-incompatible changes that occurred to charmcraft, and pins tests to use 1.6/edge rather than latest/edge.


DnPlas (approved):
LGTM

PR #115, opened 2022-12-02
fix: pin integration CI to juju agent-version=2.9.34

ca-scribner:
This is a temporary fix to address [this bug](https://bugs.launchpad.net/juju/+bug/1992833).  In particular, leaving the agent-version unpinned results in the kfp-api observability tests to fail with prometheus-k8s failing in either the stop hook or the metrics-endpoint-relation-changed hook


ca-scribner:
Integration CI for kfp-profile-controller and kfp-api should work here, but the full bundle tests will fail because of other errors.  Those other errors will be addressed in #116

DnPlas (approved):

PR #131, opened 2022-12-09
Don't scan requirements*.txt for renovate runs

misohu:
We want only to scan pip compile requirements*.in files during renovate run not requirements*.txt. This PR also tries to fix chicken and egg problem during pip compile by specifying explicit order of scans.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,13 +2,24 @@
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "rebaseWhen": "behind-base-branch",
     "dependencyDashboard": false,
+    "constraints": {
+        "python": "3.8.0"
+    },
     "pip-compile": {
         "fileMatch": [
-            "(^|/)requirements.*\\.in$"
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -2,14 +2,25 @@
...
+        "python": "3.8.0"
+    },
     "pip-compile": {
         "fileMatch": [
-            "(^|/)requirements.*\\.in$"
+            "(^|/)requirements.in",
+            "(^|/)requirements-fmt.in",
+            "(^|/)requirements-lint.in",
+            "(^|/)requirements-unit.in",
+            "(^|/)requirements-integration.in"
```
ca-scribner, line 14: I think I know why you have these all listed out and I'm :+1: to that, but can we also put the catch-all pattern of `"(^|/)requirements.*\\.in$"` at the end?  That way if someone adds another environment+requirements file, it'll *probably* have the right thing occur

i-chvets (approved):
Approved.

PR #153, opened 2023-02-08
upgrade charm to v2.0-alpha.6 for 1.7

NohaIhab:
## Summary:

- bumped image from v2.0.0-alpha.5 to v2.0.0-alpha.6
- update manifests based on https://github.com/kubeflow/manifests/tree/v1.7.0-rc.0/apps/pipeline/upstream
**note:** sync.py is kept as is since no changes were made in upstream between 1.6.1 and 1.7.0-rc.0

fix:
- added SIMULATE_CAN_CONNECT = TRUE to unit tests to get rid of warnings


charms/kfp-api/tests/unit/test_operator.py
```diff
@@ -13,6 +14,8 @@
 
 # TODO: Tests missing for config_changed and dropped/reloaded relations
 
+ops.testing.SIMULATE_CAN_CONNECT = True
```
DnPlas, line 17: What's the need for this line?

DnPlas:
Thanks @NohaIhab, I am not so sure why you picked that PR in the description to base all your changes, but it would be preferable if you base your upgrade in the [1.7.0.rc.0](https://github.com/kubeflow/manifests/tree/v1.7.0-rc.0/apps/pipeline/upstream) branch in kubeflow/manifests. Can you confirm that those manifests/versions/etc. are the same as ours?

ca-scribner (changes requested):
Most of this looks good, but I think the sync.py part needs work.  A number of the changes in this PR revert intentional differences we've introduced to depart from what upstream does.  

Are you sure that sync.py has changed between 1.6.1 and 1.7?  I don't see any changes, but I might be looking at the wrong thing

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -486,4 +487,4 @@ def do_POST(self):
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
ca-scribner, line 490:
```suggestion
    main()

```

ca-scribner (approved):
lgtm!

PR #167, opened 2023-03-15
add traceback into log

beliaev-maksim:


beliaev-maksim:
with following PR the log looks like:

```
application-kfp-api: 12:56:49 ERROR unit.kfp-api/2.juju-log object-storage:19: Encountered the following exception when parsing mysql relation: list index out of range
Traceback (most recent call last):
  File "./src/charm.py", line 300, in _get_mysql
    unit = list(mysql.units)[0]
IndexError: list index out of range
```

PR #187, opened 2023-03-30
KF-3692 replace CI with reusable workflow

beliaev-maksim:
Add reusable workflows from 
https://github.com/canonical/charmed-kubeflow-workflows/tree/all/.github/workflows

You may see that the file is quite long and could be further extracted. Unfortunately, current charms state does not allow it.
We need to have the common set of tests across all the charms, then we will be able to further simplify. However, it is already 81line with logic that should not be changed vs 334 lines of constant updates.

Required across the entire Kubeflow portfolio:
* [x] Lint
* [x] Libs
* [x] Unit
* [ ] Observability
* [ ] Bundle
* [ ] Integration (for all the charms within the repo)

Here is another example PR with similar idea: https://github.com/canonical/knative-operators/pull/95


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+    - track/**
+
+# kill all previous jobs in the same branch if new commits are pushed
```
beliaev-maksim, line 10:
```suggestion
# cancel all previous jobs in the same branch if new commits are pushed
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+      matrix:
+        charm:
+          - kfp-api
+          - kfp-persistence
+          - kfp-profile-controller
+          - kfp-schedwf
+          - kfp-ui
+          - kfp-viewer
+          - kfp-viz
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/_quality-checks.yaml@all
```
DnPlas, line 28: Why do we have an `all` branch?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+# kill all previous jobs in the same branch if new commits are pushed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  quality-checks:
+    name: Quality Checks
+    strategy:
+      matrix:
```
DnPlas, line 19: I believe the matrix can be generated from the repository. If we automate this part, we could get rid of future issues with updating this list if any directory changes name or is removed.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+    secrets: inherit
+    with:
+      charm-name: ${{ matrix.charm }}
+
+  integration-tests:
+    name: Integration Tests
+    needs: quality-checks
+    strategy:
+      matrix:
+        charm:
```
DnPlas, line 55: I think we could automate here: `if not integration dir; then skip integration tests; else run integration tests`

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+    with:
+      charm-name: ${{ matrix.charm }}
+
+  bundle-tests:
+    name: Bundle Tests
+    needs: quality-checks
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/_bundle.yaml@all
+
+  publish-on-success:
+    # only publish on push to main or track/** and if all tests pass
```
DnPlas, line 69:
```suggestion
    # only publish on push to main or track/** if all tests pass
```

.github/workflows/ci.yaml
beliaev-maksim: all branch is in canonical/charmed-kubeflow-workflows

once we finalize, I think we would need to switch to tags and V1, V2, etc

currently, I incorporate feedback directly in the branch and do not need to reapply tags

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+          - kfp-api
+          - kfp-profile-controller
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/_integration-tests.yaml@all
+    secrets: inherit
+    with:
+      charm-name: ${{ matrix.charm }}
+
+  bundle-tests:
+    name: Bundle Tests
+    needs: quality-checks
```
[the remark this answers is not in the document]
DnPlas: I don't love the name "bundle tests", though we have a `bundle.yaml` file, I don't think we promote this as a bundle. I'd rather name this "Repository Level E2E Tests" or similar, to avoid confusion with:
1) the CKF bundle
2) thinking we have a pipelines bundle

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+          - kfp-ui
+          - kfp-viewer
+          - kfp-viz
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/_unit-tests.yaml@all
+    secrets: inherit
+    with:
+      charm-name: ${{ matrix.charm }}
+
+  integration-tests:
+    name: Integration Tests
```
DnPlas, line 51: Maybe call this "Individual Integration Tests" to be more specific.

DnPlas (commented):
Thanks @beliaev-maksim, this is looking very complete so far in spite of the limitations you mention.

.github/workflows/ci.yaml
beliaev-maksim: I thought about this, but you can see that two matrices are different. That is because not all the charms have itests

one possible step for us could be create itest tox environments for all the charms, even if it is a dummy one and just always passes

DnPlas: Okay, let's keep this and improve if needed.

DnPlas (approved):
Provided the CI passes, we can merge this PR. Thanks @beliaev-maksim !

DnPlas (approved):
This PR will be merged into a development branch because we are facing issues with running a number of LXD containers at the same time for building all the kfp-operators components in different jobs. The issue has already been filed to [charmcraft](https://github.com/canonical/charmcraft/issues/1132).
To avoid blocking other PRs that depend on this as well as moving on with, the team will provide a solution that avoids running multiple LXD containers simultaneously, but that will only happen in a follow up PR.

Issue #188, opened 2023-04-12
KFP API charm rewrite using sidecar pattern


i-chvets:
# Description
KFP API charm rewrite using sidecar pattern

# Design
Design is specified in corresponding spec. 
Main design points summary:
- Use of sidecar pattern and Pebble layer.
- Porting of managing K8S resources from Pod Spec to Pebble framework.
- Use container interaction features from Charmed Operator Framework.
- Pebble layer checks use the metrics URL as HTTP endpoint.
- Handle upgrade and update status events

# Testing

i-chvets:
Completed with issues filed regarding update and remove event integration test.

PR #197, opened 2023-04-15
feat: added update-status handler

i-chvets:
https://github.com/canonical/kfp-operators/issues/188
    
Summary of changes:
- Added update-status handler.
- Agreed design with just status check and no updates to resources.
- Added unit test.


charms/kfp-api/src/charm.py
```diff
@@ -474,6 +501,14 @@ def _on_remove(self, _):
                 raise error
         self.unit.status = MaintenanceStatus("K8S resources removed")
 
+    def _on_update_status(self, _):
+        """Update status actions."""
+        self._check_leader()
```
beliaev-maksim, line 506: No way
That will put the app into error state

beliaev-maksim (changes requested):

charms/kfp-api/src/charm.py
ca-scribner: Agreed with @beliaev-maksim, I think what you want is to put the `self._check_leader()` inside the try/except.  Right now `_check_leader()` will raise an `ErrorWithStatus` if we are not the leader.  That Exception is not caught here, so it goes to Juju and sets the unit in Error rather than in `WaitingStatus("Waiting for leadership")`

If this is in the try/except (or in its own try/except, which immediately returns after setting status in the exception), then it works as expected

charms/kfp-api/src/charm.py
```diff
@@ -474,6 +501,14 @@ def _on_remove(self, _):
...
+        try:
+            self._check_status()
+        except ErrorWithStatus as err:
+            self.model.unit.status = err.status
```
ca-scribner, line 510: This is not absolutely needed for current function, but I think we should include it because the intent for this `except` block is it should set status and then exit.  There doesn't happen to be anything after so it doesn't matter now, but if someone added more code below in a future patch they might not realise check_status could error and then still continue.  

```suggestion
            self.model.unit.status = err.status
            return
```

ca-scribner (changes requested):

charms/kfp-api/src/charm.py
i-chvets: Noted. Will change.

charms/kfp-api/src/charm.py
i-chvets: That's a typo/bug.

charms/kfp-api/src/charm.py
```diff
@@ -243,6 +251,25 @@ def _check_container_connection(self, container: Container) -> None:
...
+        self._check_leader()
+        container = self.unit.get_container(self._container_name)
+        if container:
+            try:
+                check = container.get_check("kfp-api-up")
+            except ModelError as error:
+                raise GenericCharmRuntimeError(
+                    "Failed to run health check on workload container"
+                ) from error
+            if check.status != CheckStatus.UP:
```
beliaev-maksim, line 265:
```suggestion
                ) from error
                
            if check.status != CheckStatus.UP:
```

charms/kfp-api/src/charm.py
```diff
@@ -243,6 +251,25 @@ def _check_container_connection(self, container: Container) -> None:
...
+                self.logger.error(
+                    f"Container {self._container_name} failed health check. It will be restarted."
+                )
+                raise ErrorWithStatus("Workload failed health check", MaintenanceStatus)
+            else:
+                self.model.unit.status = ActiveStatus()
```
beliaev-maksim, line 271:
```suggestion
            self.model.unit.status = ActiveStatus()
```

ca-scribner (approved):
lgtm!

beliaev-maksim (dismissed):

charms/kfp-api/src/charm.py
i-chvets: This one failed linting. Removing whitespace.

ca-scribner (dismissed):

beliaev-maksim (approved):

charms/kfp-api/src/charm.py
```diff
@@ -162,9 +169,9 @@ def _kfp_api_layer(self) -> Layer:
             "checks": {
                 "kfp-api-up": {
                     "override": "replace",
-                    "period": "30s",
-                    "timeout": "20s",
-                    "threshold": 4,
+                    "period": "5m",
```
i-chvets, line 172: 5 minutes feels like a very long time for the health check. I understand that we are fixing low resources GH runners issue, but having such long period for health check is not good for real deployment. With threshold of 3 we will know that kfp-api is down after 15 minutes. Unlsess I am misreading this.

Issue #206, opened 2023-04-21
KFP API: wrong database name after MySQL integration


i-chvets:
# Description
KFP API: wrong database name is supplied to workload after MySQL integration.

Problem 1:
Previously default DB name was `mlpipeline`. MariaDB has this as configuration option. During MySQL integration, database for `mysql` relation was using new default `kfp-api-db`, when it should have taken it from relation supplied by MariaDB.

To solve the problem, need to modify code to retrieve database name from relation for `mysql`.
Changed default to `mlpipeline` to be used with `rational-db` relation to be consistent between version.

Problem 2:
Incorrect file name `config` was used.

To solve the problem, need to add `.json` extension to file being pushed.

i-chvets:
Fixed.

Issue #204, opened 2023-04-20
KFP API integration test with grafana and prometheus need to be addressed


i-chvets:
# Description
KFP API integration tests with grafana and prometheus need to be addressed.
`grafana-k8s` stuck in `unknonw` state. Test times out.
Currently, it is skipped. Needs to be reviewed and re-implemented.

Logs from CI
```
tests/integration/test_charm.py::test_prometheus_grafana_integration 
-------------------------------- live log call ---------------------------------
INFO     juju.model:model.py:2088 Deploying ch:amd64/focal/prometheus-k8s-103
INFO     juju.model:model.py:2088 Deploying ch:amd64/focal/grafana-k8s-64
INFO     juju.model:model.py:2088 Deploying ch:amd64/focal/prometheus-scrape-config-k8s-39
INFO     juju.model:model.py:2[71](https://github.com/canonical/kfp-operators/actions/runs/4756791884/jobs/8453638029#step:5:71)5 Waiting for model:
  grafana-k8s/0 [allocating] waiting: installing agent
  prometheus-k8s/0 [allocating] waiting: installing agent
  prometheus-scrape-config-k8s/0 [allocating] waiting: installing agent
INFO     juju.model:model.py:2715 Waiting for model:
  grafana-k8s/0 [allocating] waiting: agent initializing
  prometheus-k8s/0 [idle] waiting: Waiting for resource limit patch to apply
  prometheus-scrape-config-k8s/0 [executing] active: 
INFO     juju.model:model.py:2715 Waiting for model:
  grafana-k8s/0 [executing] unknown: 
. . .
  prometheus-k8s/0 [idle] waiting: Waiting for resource limit patch to apply
INFO     juju.model:model.py:2715 Waiting for model:
  grafana-k8s/0 [idle] unknown: 
  prometheus-k8s/0 [idle] blocked: Failed to apply resource limit patch: Unauthorized
INFO     juju.model:model.py:2715 Waiting for model:
  grafana-k8s/0 [idle] unknown: 
. . . 
  prometheus-k8s/0 [idle] blocked: Failed to apply resource limit patch: Unauthorized
INFO     juju.model:model.py:2715 Waiting for model:
  grafana-k8s/0 [idle] unknown:
FAILED
------------------------------ live log teardown -------------------------------
INFO     pytest_operator.plugin:plugin.py:756 Model status:

Model            Controller                Cloud/Region        Version  SLA          Timestamp
[...]

PR #207, opened 2023-04-21
Fix db name and config

i-chvets:
#206 
    
    Summary of changes:
    - Modified code to use `mlpipeline` database name as before.
    - Added type to config file pushed.

### DIFF
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -15,7 +15,7 @@ from charmed_kubeflow_chisme.exceptions import ErrorWithStatus, GenericCharmRunt
 from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
 from charmed_kubeflow_chisme.lightkube.batch import delete_many
 from charmed_kubeflow_chisme.pebble import update_layer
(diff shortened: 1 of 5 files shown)


charms/kfp-api/src/charm.py
```diff
@@ -310,8 +310,7 @@ def _upload_files_to_container(self, config_json):
             raise error
         try:
             file_content = json.dumps(config_json)
-            # no need to add `.json` extension to config file, it is detected automatically
-            config = CONFIG_DIR / "config"
+            config = CONFIG_DIR / "config.json"
```
beliaev-maksim, line 313: will it work the same for Maria?
or we do not need extension there?

i-chvets: It will work for both. That's where workload (apiserver) takes its config and then talks to the DB.
Apparently, my implementation was flawed from the beginning. apiserver used default config. And it took 10-15 minutes to timeout.
Our update-status implementation actually detected it.

beliaev-maksim (approved):

i-chvets:
CI is failing due to Grafana charm [issue](https://github.com/canonical/kfp-operators/issues/204)

PR #208, opened 2023-04-22
Fix grafana edge

i-chvets:


i-chvets:
Closing. No need to use `edge` int tests.

PR #215, opened 2023-05-01
Fix add wait block update status

i-chvets:
Summary of changes:
    - Added Waiting and Blocked status handling to update status event
      handler: do not update workload status if in Waiting or Blocked state.

NOTE: CI might fail due to Grafana issue. Tracked [here](https://github.com/canonical/kfp-operators/issues/204)

### DIFF
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -590,7 +590,7 @@ class KfpApiOperator(CharmBase):
         """Update status actions."""
         try:
             self._on_event(_)
-        except ErrorWithStatus as err:
+        except ErrorWithStatus:
             return
 
         if isinstance(self.model.unit.status, WaitingStatus) or isinstance(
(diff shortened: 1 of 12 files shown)


charms/kfp-api/src/charm.py
```diff
@@ -588,7 +588,17 @@ def _on_remove(self, _):
 
     def _on_update_status(self, _):
         """Update status actions."""
-        self._on_event(_)
+        try:
+            self._on_event(_)
+        except ErrorWithStatus as err:
+            raise err
```
beliaev-maksim, line 594: do we really raise ?

this will put the unit into error state

beliaev-maksim (changes requested):

charms/kfp-api/src/charm.py
i-chvets: I think you are correct. Should we just return or pass in this case?

i-chvets: Fixed.

beliaev-maksim (approved):

Issue #220, opened 2023-06-14
KFP API: Pebble service fails to start


phoevos:
We're bumping into this issue intermittently on a clean install with the `2.0/edge` (rev 413) version of the charm, when deploying the Kubeflow bundle (`1.7/edge`) either on MicroK8s or Charmed Kubernetes.

On startup, the KFP API Server tries to connect to MinIO. If MinIO is not yet available, the connection fails with the following error, causing the service to crash, less than 1 second after it started.
```
Failed to check if Minio bucket exists. Error: Get "http://minio.kubeflow:9000/mlpipeline/?location=": dial tcp 10.152.183.92:9000: connect: connection refused
```
Due to the service failing fast (<1 sec), Pebble considers it to be inactive:
```bash
$ pebble services       
Service                 Startup  Current   Since
ml-pipeline-api-server  enabled  inactive  -
```
This is a design decision on the Pebble side as explained [here](https://github.com/canonical/pebble#viewing-starting-and-stopping-services):
> When starting a service, Pebble executes the service's command, and waits 1 second to ensure the command doesn't exit too quickly. Assuming the command doesn't exit within that time window, the start is considered successful, otherwise pebble start will exit with an error.

Due to the fact that the service was never active, Pebble never attempts to restart it, despite the failing health checks.

### Workaround

Since this issue is exposed due to a race (i.e. MinIO not yet available), it won't come up every time. If it does occur during deployment (after the rest of the bundle has been installed successfully), however, we need to start the API Server service manually to unblock:
```
juju ssh kfp-api/0 "PEBBLE_SOCKET=/charm/containers/ml-pipeline-api-server/pebble.socket /charm/bin/pebble replan"
```

### Mitigation

We need to come up with the plan to avoid bumping into this issue in the future. There's a couple of things that could be done on our side:
* Sleep for 1 second before starting any Pebble service
  * This is not ideal, since it defeats the purpose of this Pebble feature 
* Catch the error and manually retry starting the service on the charm side
  * This is not ideal, since it shifts the responsibility of service management from the service manager itself to the high-level charm code

DnPlas:
I also hit this issue today.

#### My environment
* Ubuntu 20.04
* juju controller 2.9.34
* kfp-api 2.0/stable rev413
* minio ckf-1.7/stable rev186

#### Steps to reproduce
[...]

PR #221, opened 2023-06-20
pebble: Sleep before starting KFP API server

phoevos:
Sleep for 1 second before attempting to start the KFP API server, in order to avoid issues with Pebble when the corresponding service fails fast. This is a temporary measure until a fix for canonical/pebble#240 is provided.

Refs #220

### DIFF
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2021 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 """Charm the Kubeflow Pipelines API.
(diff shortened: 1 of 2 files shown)


charms/kfp-api/src/charm.py
```diff
@@ -187,7 +187,7 @@ def _kfp_api_layer(self) -> Layer:
                 self._container_name: {
                     "override": "replace",
                     "summary": "ML Pipeline API Server",
-                    "command": self._exec_command,
+                    "command": f"bash -c '{self._exec_command}'",
```
beliaev-maksim, line 190: no sleep?

beliaev-maksim (changes requested):

charms/kfp-api/src/charm.py
beliaev-maksim: I was looking at only one commit....

beliaev-maksim (approved):
LGTM if CI passes

beliaev-maksim (approved):

PR #232, opened 2023-06-28
tests: simplify and extract bundle management logic

DnPlas:
This commit simplifies the bundle rendering and applying logic, and abstracts it from the test files so it can be reused. This reusable set of helper methods leverage pytest-operator's render_bundle()[1] and COS' cli_deploy_bundle()[2].

Links:
[1] https://github.com/charmed-kubernetes/pytest-operator/blob/main/pytest_operator/plugin.py#L1232 [2] https://github.com/canonical/cos-lite-bundle/blob/main/tests/integration/helpers.py#L19


tests/integration/data/bundle.yaml.j2
```diff
@@ -0,0 +1,55 @@
+bundle: kubernetes
+name: kubeflow-pipelines
+applications:
+  argo-controller:         { charm: ch:argo-controller, channel: latest/edge, scale: 1 }
+  metacontroller-operator: { charm: ch:metacontroller-operator, channel: latest/edge, scale: 1, trust: true }
+  minio:                   { charm: ch:minio, channel: latest/edge,       scale: 1 }
+  kfp-db:                  { charm: charmed-osm-mariadb-k8s, channel: latest/stable, scale: 1, options: { database: mlpipeline } }
```
kimwnasptd, line 7: Can we get this bundle as close as possible to what's used right now in the Kubeflow bundle for KFP? Talking about MariaDB here.

Ideally I'd say we should just reference this bundle from the Kubeflow bundle

DnPlas: Agreed, I can update it to replace mariadb with mysql-k8s.

> Ideally I'd say we should just reference this bundle from the Kubeflow bundle

That will slightly change the placement of this bundle file, but we can definitely do that. Do you mind if I add that as a secondary task after we merge this PR?

kimwnasptd: Sure!

DnPlas: I have opened an [issue](https://github.com/canonical/kfp-operators/issues/233) for discussing this. Thanks for the suggestion @kimwnasptd

DnPlas:
Closing in favour of #238

PR #247, opened 2023-07-12
build: use charmedkubeflow container images

DnPlas:
Use container images from the charmedkubeflow container image registry. The affected charms are: api, persistence, schedwf, and viewer.

This PR also includes a change for renaming the service name of the KFP API server in the `kfp-api` charm code. For convenience and to avoid running (or attempting to run two services), the service name in the charm code should match that of the Rockcraft project [here](https://github.com/canonical/pipelines-rocks/blob/main/api-server/rockcraft.yaml#L11). The charm code will anyway replace it with its own configuration. This also ensures the changes made to the Pebble layer in the charm code are taken into account, for instance [this](https://github.com/canonical/kfp-operators/commit/06303e4f0d2284b532758fd4f30499898888b797), which is a fix for https://github.com/canonical/pipelines-rocks/issues/22.
Fixes: https://github.com/canonical/pipelines-rocks/issues/22

Blocked by #250, merge after #261


charms/kfp-api/src/charm.py
```diff
@@ -180,11 +180,13 @@ def service_environment(self):
     @property
     def _kfp_api_layer(self) -> Layer:
         """Create and return Pebble framework layer."""
+        # The service name should be the same as the one
+        # defined in the Rockcraft project: apiserver
         layer_config = {
             "summary": "kfp-api layer",
             "description": "Pebble config layer for kfp-api",
             "services": {
-                self._container_name: {
```
beliaev-maksim, line 187: why do we change it in only one place?

DnPlas: What do you mean only in one place?

beliaev-maksim: you changed attributed to hardcoded value. I am curious if that should be done in other places as well

beliaev-maksim (dismissed):

i-chvets (approved):
Approved.

PR #249, opened 2023-07-13
tests: remove wait for deployments workaround

DnPlas:
Removing the workaround as the proposed fix in [1] should take care of it.

Links:
[1] https://github.com/juju/python-libjuju/pull/905


cderici:
@DnPlas Looks like the test failed `juju/model.py, line 2514, in _raise_for_status`, I'm on it 👍

DnPlas:
> > @DnPlas Looks like the test failed `juju/model.py, line 2514, in _raise_for_status`, I'm on it 👍
> 
> Well looks like it's raised because the unit is actually in error (`Unit in error: kfp-ui/0`), can you confirm that this is because of `wait_for_idle`'s behavior? If so, is this the same scenario in [juju/python-libjuju#900](https://github.com/juju/python-libjuju/issues/900) ?

Hi @cderici , it seems like an unrelated issue. The kfp-ui pod is not scheduled due to the GH runner running out of storage (?). This just happened today, we are investigating. In the meantime for testing the fix, I will try in a bigger machine.

DnPlas:
Update: https://github.com/juju/python-libjuju/pull/905#issuecomment-1635672596

DnPlas:
Closing, this PR has served its purpose.

Issue #250, opened 2023-07-13
`The node was low on resource: ephemeral-storage.` message when running bundle integration tests


DnPlas:
It seems like we are hitting a storage limit when deploying the `kfp-operators` on GitHub runners. While executing the bundle integration tests, some pods have this message in their events, for example:

* kfp-persistence
```
Name:         kfp-persistence-5fc5568559-thtnb
Namespace:    kubeflow
Priority:     0
Node:         fv-az621-251/10.1.0.19
Start Time:   Thu, 13 Jul 2023 14:05:45 +0000
Labels:       app.kubernetes.io/name=kfp-persistence
              pod-template-hash=5fc5568559
Annotations:  apparmor.security.beta.kubernetes.io/pod: runtime/default
              charm.juju.is/modified-version: 0
              cni.projectcalico.org/containerID: 5634b65404f070290adb448420e37da53529093ee73a7c4b74ee36e21102d3da
              cni.projectcalico.org/podIP: 
              cni.projectcalico.org/podIPs: 
              controller.juju.is/id: 1b2ee3dc-da12-4a98-8001-cc0d5ac2c978
              model.juju.is/id: 6b99181f-5a33-4a55-882f-b373cb244951
              seccomp.security.beta.kubernetes.io/pod: docker/default
              unit.juju.is/id: kfp-persistence/0
Status:       Failed
Reason:       Evicted
Message:      The node was low on resource: ephemeral-storage. 
```

* kfp-ui
```
Events:
  Type     Reason               Age   From     Message
  ----     ------               ----  ----     -------
  Warning  Evicted              20s   kubelet  The node was low on resource: ephemeral-storage. Container juju-operator was using 265896Ki, which exceeds its request of 0.
  Normal   Killing              20s   kubelet  Stopping container juju-operator
  Warning  ExceededGracePeriod  10s   kubelet  Container runtime did not kill the pod within specified grace period.
```

The last two CI runs fail with because of this: [example 1](https://github.com/canonical/kfp-operators/actions/runs/5544299869) and [example 2](https://github.com/canonical/kfp-operators/actions/runs/5542344265)

PR #255, opened 2023-07-17
fix: add ml-pipeline-service.yaml.j2 in K8S_RESOURCE_FILES

DnPlas:
The `ml-pipeline-service.yaml.j2` template was not being rendered and applied as it was not listed in `K8S_RESOURCE_FILES`. This template file contains the `ml-pipeline Service`, needed by the KFP API.

This change should be backported to `track/2.0`.

> NOTE: the bundle tests will most likely fail because of #250, merge after #261


i-chvets (approved):
Approved.

Issue #243, opened 2023-07-10
`kfp-api` individual integration tests fail (potential race condition)


DnPlas:
`kfp-api` [individual integration tests](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/tests/integration/test_charm.py) fail due to `kfp-api` going into error state with the following message:

```
unit-kfp-api-0: 07:29:37 ERROR unit.kfp-api/0.juju-log Failed to generate container configuration.
unit-kfp-api-0: 07:29:37 ERROR unit.kfp-api/0.juju-log Failed to handle <LeaderElectedEvent via KfpApiOperator/on/leader_elected[4]> with error: Please add required database relation: eg. relational-db
unit-kfp-api-0: 07:29:39 ERROR unit.kfp-api/0.juju-log Failed to generate container configuration.
unit-kfp-api-0: 07:29:39 ERROR unit.kfp-api/0.juju-log Failed to handle <PebbleReadyEvent via KfpApiOperator/on/ml_pipeline_api_server_pebble_ready[7]> with error: Please add required database relation: eg. relational-db
unit-kfp-api-0: 07:29:41 ERROR unit.kfp-api/0.juju-log Failed to generate container configuration.
unit-kfp-api-0: 07:29:41 ERROR unit.kfp-api/0.juju-log Failed to handle <ConfigChangedEvent via KfpApiOperator/on/config_changed[10]> with error: Please add required database relation: eg. relational-db
unit-kfp-api-0: 07:34:04 ERROR unit.kfp-api/0.juju-log Failed to generate container configuration.
unit-kfp-api-0: 07:34:04 ERROR unit.kfp-api/0.juju-log Failed to handle <UpdateStatusEvent via KfpApiOperator/on/update_status[16]> with error: Please add required database relation: eg. relational-db
```

For a full log go [here](https://github.com/canonical/kfp-operators/actions/runs/5505194002/jobs/10033719818#step:5:269). 

This happens intermittently in the GitHub CI when the charm is deployed and a `wait_for_idle()` call is made, but we cannot discard an issue on a regular deployment in production.

DnPlas:
It seems like this issue and [this other one](https://github.com/canonical/kfp-operators/issues/243#issuecomment-1647645819) are caused by a series of factors, but mainly because of the units in the model not being active and idle when certain test cases run. The following describes these factors:

[...]

PR #263, opened 2023-07-20
build: install setuptools and pip in charms that are missing it

DnPlas:
Installing these missing python packages so the charm gets the latest version of `pip` and `setuptools`.


orfeas-k (approved):

DnPlas:
I merge this PR as is, the kfp-api individual integration test is failing because of #243 and should be fixed soon. Merging this PR will unblock other contributions.

PR #269, opened 2023-07-24
Update the bug template

kimwnasptd:
Update the bug template, based on o11y team.


DnPlas:
@kimwnasptd I believe the CI is failing because of #243, which is something we are going to be fixing this week.

i-chvets (approved):
Approved. ACK. Create follow-up on this.

i-chvets (approved):
Approved with ACK as noted in comments.

PR #272, opened 2023-07-26
fix: Handle relation-broken events

### LINKED ISSUE #222: error on relational-db-relation-broken
I removed kfp-api application from bundle and kfp went into error state

eg
```
juju remove-application kfp-api
```

```
2023-06-20T10:02:41.398Z [container-agent] 2023-06-20 10:02:41 DEBUG juju-log relational-db:18: Operator Framework 2.3.0 up and running.
2023-06-20T10:02:41.420Z [container-agent] 2023-06-20 10:02:41 DEBUG juju-log relational-db:18: Emitting Juju event relational_db_relation_broken.
 Error in sys.excepthook:
 Traceback (most recent call last):
   File "/usr/lib/python3.8/logging/__init__.py", line 954, in handle
     self.emit(record)
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/log.py", line 41, in emit
     self.model_backend.juju_log(record.levelname, self.format(record))
   File "/usr/lib/python3.8/logging/__init__.py", line 929, in format
     return fmt.format(record)
   File "/usr/lib/python3.8/logging/__init__.py", line 676, in format
     record.exc_text = self.formatException(record.exc_info)
   File "/usr/lib/python3.8/logging/__init__.py", line 626, in formatException
     traceback.print_exception(ei[0], ei[1], tb, None, sio)
   File "/usr/lib/python3.8/traceback.py", line 103, in print_exception
     for line in TracebackException(
   File "/usr/lib/python3.8/traceback.py", line 612, in format
     yield from self.__context__.format(chain=chain)
   File "/usr/lib/python3.8/traceback.py", line 617, in format
     yield from self.format_exception_only()
   File "/usr/lib/python3.8/traceback.py", line 566, in format_exception_only
     stype = smod + '.' + stype
 TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
 
 Original exception was:
 Traceback (most recent call last):
   File "./src/charm.py", line 517, in _get_db_data
     db_data = self._get_mysql_data()
   File "./src/charm.py", line 446, in _get_mysql_data
     relation = self._get_db_relation("mysql")
   File "./src/charm.py", line 436, in _get_db_relation
     raise GenericCharmRuntimeError(
 <unknown>GenericCharmRuntimeError: Database relation mysql is not established or empty
 
 During handling of the above exception, another exception occurred:
 
 Traceback (most recent call last):
   File "./src/charm.py", line 683, in <module>
     main(KfpApiOperator)
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/main.py", line 441, in main
     _emit_charm_event(charm, dispatcher.event_name)
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/main.py", line 149, in _emit_charm_event
     event_to_emit.emit(*args, **kwargs)
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/framework.py", line 354, in emit
     framework._emit(event)
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/framework.py", line 830, in _emit
     self._reemit(event_path)
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/framework.py", line 919, in _reemit
     custom_handler(event)
   File "./src/charm.py", line 662, in _on_relational_db_relation_remove
     self._on_event(event)
   File "./src/charm.py", line 669, in _on_event
     config_json = self._generate_config(interfaces)
   File "./src/charm.py", line 217, in _generate_config
     db_data = self._get_db_data()
   File "./src/charm.py", line 524, in _get_db_data
     db_data = self._get_relational_db_data()
   File "./src/charm.py", line 482, in _get_relational_db_data
     relation_data = self.database.fetch_relation_data()
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/lib/charms/data_platform_libs/v0/data_interfaces.py", line 508, in fetch_relation_data
     key: value for key, value in relation.data[relation.app].items() if key != "data"
   File "/var/lib/juju/agents/unit-kfp-api-0/charm/venv/ops/model.py", line 1346, in __getitem__
     raise KeyError(
 KeyError: 'Cannot index relation data with "None". Are you trying to access remote app data during a relation-broken event? This is not allowed.'

```

phoevos:
In the kfp-api charm, `*-relation-broken` events trigger the same workflow as every other event, forcing the charm to recalculate its configuration. In order to do that, it attempts to retrieve info from its required relations. When executing in response to a relation-broken event for a removed `relational-db` relation, however, the library call to fetch the data raises an error. We need to catch this error and properly propagate it so that the charm lands in the desired state (i.e. Blocked, waiting for the required relation to be added).

Closes #222

### DIFF
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -486,7 +486,15 @@ class KfpApiOperator(CharmBase):
         self._get_db_relation("relational-db")
 
         # retrieve database data from library
-        relation_data = self.database.fetch_relation_data()
+        try:
+            # if called in response to a '*-relation-broken' event, this will raise an exception
+            relation_data = self.database.fetch_relation_data()
+        except KeyError:
+            self.logger.error("Failed to retrieve relation data from library")
+            raise GenericCharmRuntimeError(
+                "Failed to retrieve relational-db data. This is to be expected if executed in"
+                " response to a '*-relation-broken' event"
+            )
         # parse data in relation
         # this also validates expected data by means of KeyError exception
         for val in relation_data.values():
--- a/charms/kfp-api/tests/unit/test_operator.py
+++ b/charms/kfp-api/tests/unit/test_operator.py
@@ -6,7 +6,7 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 import yaml
-from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 from ops.testing import Harness
 
 from charm import KFP_API_SERVICE_NAME, ErrorWithStatus, KfpApiOperator
@@ -572,3 +572,39 @@ class TestCharm:
                 "db_host": "host",
                 "db_port": "1234",
             }
+
+    def test_relational_db_relation_broken(
+        self,
+        mocked_resource_handler,
+        mocked_lightkube_client,
+        mocked_kubernetes_service_patcher,
+        harness: Harness,
+    ):
+        """Test that a relation broken event is properly handled."""
+        database = MagicMock()
+        fetch_relation_data = MagicMock(side_effect=KeyError())
+        database.fetch_relation_data = fetch_relation_data
+        harness.model.get_relation = MagicMock(
+            side_effect=self._get_relation_db_only_side_effect_func
+        )
+
+        rel_name = "relational-db"
+        rel_id = harness.add_relation(rel_name, "relational-db-provider")
+
+        harness.begin()
+        harness.set_leader(True)
+        harness.container_pebble_ready(KFP_API_CONTAINER_NAME)
+
+        assert harness.model.unit.status == WaitingStatus("Waiting for relational-db data")
+
+        harness.charm.database = database
+        del harness.model.get_relation
+
+        harness._emit_relation_broken(rel_name, rel_id, "kfp-api")
+
+        assert harness.model.unit.status == BlockedStatus(
+            "Please add required database relation: eg. relational-db"
+        )
+
+        harness.charm.on.remove.emit()
+        assert harness.model.unit.status == MaintenanceStatus("K8S resources removed")
--- a/charms/kfp-api/tests/integration/test_charm.py
+++ b/charms/kfp-api/tests/integration/test_charm.py
@@ -174,6 +174,9 @@ class TestCharm:
         )
         assert ops_test.model.applications[APP_NAME].units[0].workload_status == "blocked"
 
+        # remove redundant relation
+        await ops_test.juju("remove-relation", f"{APP_NAME}:mysql", "kfp-db:mysql")
+
     async def test_prometheus_grafana_integration(self, ops_test: OpsTest):
         """Deploy prometheus, grafana and required relations, then test the metrics."""
         prometheus = "prometheus-k8s"
@@ -240,3 +243,8 @@ class TestCharm:
         wait=wait_exponential(multiplier=1, min=1, max=10),
         reraise=True,
     )
+
+    async def test_remove_application(self, ops_test: OpsTest):
+        """Test that the application can be removed successfully."""
+        await ops_test.model.remove_application(app_name=APP_NAME, block_until_done=True)
+        assert APP_NAME not in ops_test.model.applications


orfeas-k (commented):
Thank you @phoevos , the changes look good. Would it make sense to add a test case in integration tests for this in order to make sure we do not stumble on it again? I think we should but I don't know if this should be part of this PR. Other than that, LGTM!

charms/kfp-api/src/charm.py
```diff
@@ -486,7 +486,15 @@ def _get_relational_db_data(self) -> dict:
         self._get_db_relation("relational-db")
 
         # retrieve database data from library
-        relation_data = self.database.fetch_relation_data()
+        try:
+            # if called in response to a '*-relation-broken' event, this will raise an exception
+            relation_data = self.database.fetch_relation_data()
+        except KeyError:
```
DnPlas, line 492: To understand this change, could you please explain how and under which circumstances the KeyError will be raised when calling `self.database.fetch_relation_data()`?
I don't really see anything [here](https://github.com/canonical/kfp-operators/blob/0580c89e0f4143c6e781fcc4ec6ae384675c08ce/charms/kfp-api/lib/charms/data_platform_libs/v0/data_interfaces.py#L511) that could potentially raise this exception.

DnPlas (commented):
Thanks @phoevos, just a comment to understand the change.

charms/kfp-api/src/charm.py
phoevos: So, they say in the docstring that:
> Function cannot be used in `*-relation-broken` events and will raise an exception.

In their code, they do the following:
```python
relation.data[relation.app]
```

Apparently, on `relation-broken` `relation.app` can be `None`, which will raise a `KeyError` if used to index the dict with.

This can be seen in the trace provided in the linked issue: https://github.com/canonical/kfp-operators/issues/222

phoevos: Now, note, I haven't been able to reproduce this with Juju 2.9.44. I'm not sure if that's still the expected behaviour (i.e. `relation.app` becoming `None` when the relation breaks), so I implemented this based on the trace I got.

DnPlas: (from an offline conversation) I think it is okay to handle this exception here as the library doesn't for us (we should probably raise an issue in the library so they handle this better?), but IMHO to fix the issue we should be paying more attention to this:

* The error suggests that there was something wrong when handling the relational-db relation broken event`2023-06-20T10:02:41.420Z [container-agent] 2023-06-20 10:02:41 DEBUG juju-log relational-db:18: Emitting Juju event relational_db_relation_broken.`
* That event's handler is [this](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L666) 👿 
* The above will put the unit to maintenance and then call `_on_event()` which will do ALL the things that are done regularly (get interfaces, generate config, upload files to container, etc.)
* Because this is a relation broken event, some of the data that should be in the relation is not there, and thus some of these calls in `on_event()` will fail, specially [`config_json = self._generate_config(interfaces)`](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L676) because that one will call [`get_db_data()`](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L517), which also calls [this](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L531), which will eventually call [`fetch_relation_data`](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L489), the method that cannot be used in relation broken event by design.

I think for instance [this](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L489) comment is more or less valid and I'd probably look at [this](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/charm.py#L643) handler and try to make it better to handle the case of relation broken.

phoevos: The description above is accurate. I'm not sure about the last 2 links, don't think they point to the intended thing.

Currently, we use a single handler to tackle both the `relation-departed` and `relation-broken` events. Now, I believe that `relation-departed` is handled properly. We go through the same actions we do for most events, and generating the configuration will clear the fields that correspond to the departed relation. So that is actually fine.

What do we really expect the charm to do on a `relation-broken` event, though? Given that it always comes after the departure has been handled, the corresponding relation databag fields have already been cleared. This is all about the config anyway. These relations do not affect the K8s resources applied whatsoever. So what our handler will do with the proposed fix is realise that the relation is no more available and set the charm to `Blocked`, with the `"Please add required database relation: eg. relational-db"` message.

Do you have a different behaviour in mind? Is there something else that we should be doing on `relation-broken` events?

DnPlas: Sorry about the links, I have updated them in my previous comment.

> What do we really expect the charm to do on a relation-broken event, though?

I guess this is the important question here. This topic is something we have been discussing in the team for a while. I remember we had this conversation when @ca-scribner was designing the [dashboard library](https://github.com/canonical/kubeflow-dashboard-operator/blob/main/lib/charms/kubeflow_dashboard/v0/kubeflow_dashboard_links.py). Though it is different than this problem, I think it provides valuable information, and you can definitely take a look at it for inspiration.
To be honest, I would have to think on what should be done in the relation broken event handler to provide a good answer.

> These relations do not affect the K8s resources applied whatsoever. So what our handler will do with the proposed fix is realise that the relation is no more available and set the charm to Blocked, with the "Please add required database relation: eg. relational-db" message.
Do you have a different behaviour in mind? Is there something else that we should be doing on relation-broken events?

It is true the relations (parting or being added) do not affect the K8s resources applied (that is why I said the comment was more or less accurate).

Now, deleting the application with `juju remove kfp-api` is triggering a relation broken event (?), which will then call `on_event()`, calling this will cause the same actions we are doing on install or on config change to be run. This is not desired because on remove you want to remove things, not render + apply + generate config (one of the methods that is raising) + fetch relation data (one of the methods that is raising), etc.

While your proposed solution will fix the issue in the case where you remove the DB application (`mysql-k8s`) and there is a sudden relation broken, it will put the unit to BlockedStatus on remove, something that wouldn't make sense because you want to remove, not relate it to another application.

phoevos: Oh I get what you're saying.

I'm hesitant about refactoring the remove handler as part of this effort, because then we would also have to update the `joined`/`changed` handlers which don't need to apply K8s resources either, and probably others as well, since now all of them call `_on_event`.

However, we could try setting the `db_data` to something empty after catching the error instead of setting the unit in Blocked status.

I think that what you're describing though is mostly a matter of semantics. Yes, it doesn't make sense for `kfp-api` to go into `Blocked` state asking for relations when we try to remove it. However, if we deal with the raised `KeyError`, this would never actually cause an issue, being just a transient state (and most likely imperceptible on the user side). I mean, Juju will update the relation first, firing the relevant events (i.e. `departed` -> `broken`), but then it will actually proceed with the application removal, which means that the `_on_remove` handler will run at the end, deleting any K8s resources created.

phoevos: So yes, I'm all for restructuring our handlers so that they're more fine-grained and perform the exact tasks that should concern them. However, IMO this is a much greater effort design and implementation-wise, given that I'm not sure any of our charms actually go for that approach.

Taking a look at the dashboard library.

DnPlas (dismissed):

charms/kfp-api/tests/unit/test_operator.py
```diff
@@ -572,3 +572,39 @@ def test_relational_db_relation_with_data(
...
+        harness.container_pebble_ready(KFP_API_CONTAINER_NAME)
+
+        assert harness.model.unit.status == WaitingStatus("Waiting for relational-db data")
+
+        harness.charm.database = database
+        del harness.model.get_relation
+
+        harness._emit_relation_broken(rel_name, rel_id, "kfp-api")
+
+        assert harness.model.unit.status == BlockedStatus(
```
DnPlas, line 605: Just to confirm, in all the sequence of exception handling, this is the last thing that will happen, right?

DnPlas (approved):
Just a tiny question that would help me understand the unit test, if you could resolve it before merging, it would be great. LGTM, thanks @phoevos

charms/kfp-api/src/charm.py
DnPlas: @phoevos has provided with unit and integration tests that can help confirm the exception will be handled and won't interfere in the removal process. resolving this conversation.

charms/kfp-api/tests/unit/test_operator.py
phoevos: Exactly!

PR #293, opened 2023-08-04
ci, build: use action to free space from GH runners and update dependencies

DnPlas:
The action will remove unused software from the GH runners to free some storage space in /dev/root. This will allow the heavy workload that are functional tests to run properly without storage limitations. 

This commit is based on f83bba0

Related to #250


i-chvets (approved):
Approved.

PR #308, opened 2023-08-24
fix: profile-controller integration tests wait_for_idle

NohaIhab:
The `kfp-profile-controller` integration tests are sometimes failing, due to `wait_for_idle` being set to wait for the `kfp-profile-controller` charm only. This means other charms might not be ready when the `test_build_and_deploy` is done, making this test incorrect, and causing other tests to fail.
Fixed by removing the `apps=[CHARM_NAME]` in `wait_fot_idle`, so it waits for all apps.


DnPlas (approved):
Thanks @NohaIhab

Issue #323, opened 2023-09-11
KFP run stuck in None state


phoevos:
### Bug Description

The status of submitted KFP runs is never updated and is therefore stuck to `None` in the `latest/edge` version of the KFP charms. 

The Argo Workflow is submitted properly and executes successfully, but that doesn't reflect on the run itself, as seen using either the KFP client (which returns a finished run with a None status) or the KFP UI (which hangs loading).

This is likely a bug introduced in our recent sidecar rewrites. At the moment it's just a guess, but I'm thinking that this has something to do with the KFP Persistence Agent not working properly, given that the content of the KFP MySQL DB is never updated with the completed workflow.

### To Reproduce

1. Deploy KFP (it's easier to do so as part of the CKF bundle to facilitate debugging)
2. Run one of the example pipelines through the KFP UI
3. Note that it's status never changes (either using the client or navigating to the runs page and attempting to fetch the created run, which will never load)
4. Inspect the contents of `kfp-db` and verify that the workflow saved as part of the created run entry does not have an updated statusits

### Environment

* KFP charms: `latest/edge`
* MicroK8s: `1.24/stable`
* Juju: `2.9/stable`

### Relevant log output

```shell
N/A
```

### Additional context

I noticed that before the rewrite we were applying this ServiceAccount to the PersistenceAgent container, which allowed for accessing the Argo Workflow K8s resources: https://github.com/canonical/kfp-operators/blob/67235bfa402fb4f67c30521fad431c467a1b0d44/charms/kfp-persistence/src/charm.py#L62-L85

It doesn't look like we're currently applying these elsewhere. This shouldn't be an issue here, since we're deploying the charm with trust, but we should also make a note of deploying the required upstream ClusterRole and Binding.

phoevos:
### Explanation

As part of the charm rewrite we changed the value of the `--namespace` option provided when starting the persistence agent service from `""` to match the model namespace value. The config is set up here:
https://github.com/canonical/kfp-operators/blob/5806a6be8b0ca4111e33b9077ee1c245acbbdc01/charms/kfp-persistence/src/charm.py#L68
And used here:
https://github.com/canonical/kfp-operators/blob/5806a6be8b0ca4111e33b9077ee1c245acbbdc01/charms/kfp-persistence/src/components/pebble_components.py#L54

[...]

PR #325, opened 2023-09-12
fix: Unset namespace in persistence agent config

phoevos:
Currently, we provide the --namespace option when starting the persistence agent, setting its value to match that of the model where the charm is deployed. However, this won't work for multi-user Kubeflow installations, where the namespace value should be empty as per the corresponding upstream [manifests](https://github.com/kubeflow/manifests/blob/92542398b2efbbfd9498e40a7cef6e94c68430bf/apps/pipeline/upstream/base/installs/multi-user/persistence-agent/deployment-patch.yaml#L14-L15).

Closes https://github.com/canonical/kfp-operators/issues/323


DnPlas (approved):
Thanks @phoevos LGTM, as mentioned in the linked issue, it was a regression introduced in a recent refactor.

PR #333, opened 2023-09-21
fix: updated workflows

i-chvets:
Summary of changes:
- Updated GH workflows with 22.04


DnPlas:
Thanks @i-chvets, this change will not be required as we are still using 20.04, plus this branch will soon fall out of support. Closing this PR.

Issue #335, opened 2023-09-25
Make charm's images configurable in track/<last-version> branch


DnPlas:
## Description

The goal of this task is to make all images configurable so that when this charm is deployed in an airgapped environment, all image resources are pulled from an arbitrary local container image registry (avoiding pulling images from the internet).
This serves as a tracking issue for the required changes and backports to the latest stable `track/*` Github branch.

### TL;DR
Mark the following as done

- [ ] Required changes (in metadata.yaml, config.yaml, src/charm.py)
- [ ] Test on airgap environment
- [ ] Publish to <version>/stable

## Required changes

> WARNING: No breaking changes should be backported into the `track/<version>` branch. A breaking change can be anything that requires extra steps to refresh from the previous <version>/stable other than just `juju refresh`. Please avoid at all costs these situations.

The following files have to be modified and/or verified to enable image configuration:

* `metadata.yaml` - the container image(s) of the workload containers have to be specified in this file. This only applies to sidecar charms. Example:

```yaml
containers:
  training-operator:
    resource: training-operator-image
resources:
  training-operator-image:
    type: oci-image
    description: OCI image for training-operator
    upstream-source: kubeflow/training-operator:v1-855e096
```

* config.yaml - in case the charm deploys containers that are used by resource(s) the operator creates. Example:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: seldon-config
  namespace: {{ namespace }}
data:
  predictor_servers: |-
    {
        "TENSORFLOW_SERVER": {
          "protocols" : {
            "tensorflow": {
              "image": "tensorflow/serving", <--- this image should be configurable
              "defaultImageVersion": "2.1.0"
              },
            "seldon": {
              "image": "seldonio/tfserving-proxy",
              "defaultImageVersion": "1.15.0"
              }
            }
        },
...
```

* tools/get-images.sh - is a bash script that returns a list of all the images that are used by this charm. In the case of a multi-charm repo, this is located at the root of the repo and gathers images from all charms in it.

* src/charm.py - verify that nothing inside the charm code is calling a subprocess that requires internet connection.

## Testing

1. Spin up an airgap environment following https://github.com/canonical/bundle-kubeflow/pull/682 and https://github.com/canonical/bundle-kubeflow/issues/703#issuecomment-1729826733

2. Build the charm making sure that all the changes for airgap are in place.

3. Deploy the charms manually and observe the charm go to active and idle.

[...]

PR #336, opened 2023-09-26
scripts(get-images): Include cache-image config

orfeas-k:
Modify `get-images.sh` script in order to include cache-image from config.yaml file.
Refs #335


i-chvets (approved):
Approved.

Issue #334, opened 2023-09-22
Rewrite kfp-profile-controller to use Decorator controller


misohu:
### Bug Description

Right now kfp-profile-controller uses `CompositeController` as it is written for `metacontroller-operator` image version `v2.0.4`. This causes problem to `resource-dispatcher` as it uses `DecoratroController` written for `metacontroller-operator` image version `v3.0.0+` ([issue](https://github.com/canonical/resource-dispatcher/issues/30)). 

We should rewrite kfp-profile-controller to use  `DecoratroController` so we can update `metacontroller-operator` image version to `v3.0.0+`. More about why do we have to change to `DecoratroController` is described in this [closed issue](https://github.com/canonical/metacontroller-operator/issues/84).

In order to use `DecoratroController` in kfp-profile-controller we need to rewrite `files/upstream/sync.py` file. We can take inspiration from [resource-dispatcher-image source code](https://github.com/canonical/resource-dispatcher-image/tree/main/resource_dispatcher). 

### To Reproduce

None

### Environment

None

### Relevant log output

```shell
None
```

misohu:
Fixed in https://github.com/canonical/kfp-operators/pull/342. 

**Ideas for improvements**
Currently we use sync.py file to specify the webhook's API. we push this file to the charm which makes it problematic:
- we have everything in one file
- we cannot have separate dependencies for the sync.py itself as it uses charm's one all the time
- we are not testing the API with unit tests we just relly on integration tests for charm 

The profile-controller itself has the same structure as  [resource-dispatcher ](https://github.com/canonical/resource-dispatcher). In there we decided to create a [separate image](https://github.com/canonical/resource-dispatcher-image) for the resource dispatcher api, instead of pushing file to the charm. This allow us to resolve all the problems above. We can use the same approach fot the profiles controller. We need to keep in mind that upstream is organizing code in [single sync.py file](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py) tho.

NohaIhab:
closed by #344

PR #342, opened 2023-10-02
Decorator controller rewrite

misohu:
https://github.com/canonical/kfp-operators/issues/334

Changes: 
- use Decorator controller instead of Composite controller.
- use metacontroller v3.0.0 (as it supports Decorator controller).

Tests:
- integration tests are checking also scenario where minio credentials are changed. 

This PR resolves [resource-dispatcher's issue ](https://github.com/canonical/resource-dispatcher/issues/30) as we can migrate to metacontroller v3

### DIFF
--- a/charms/kfp-profile-controller/files/upstream/sync.py
+++ b/charms/kfp-profile-controller/files/upstream/sync.py
@@ -144,7 +144,7 @@ def server_factory(visualization_server_image,
     Returns an HTTPServer populated with Handler with customized settings
     """
     class Controller(BaseHTTPRequestHandler):
-        def sync(self, parent, children):
+        def sync(self, parent, attachments):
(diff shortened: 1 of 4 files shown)


DnPlas (commented):
Thanks @misohu. I understand that this change has to be done in order to fix other (unrelated) issues, and that makes sense, but I wonder what does this change mean for the kfp-profile-controller. This is just for understanding how the operator may be affected.

misohu:
Hi @DnPlas this should not affect the controller functionality vise. It just enables us to update the metacontroller to higher version. I dont think that the operator will be affected.

DnPlas (approved):

DnPlas:
Closing as a duplicate of #344

PR #344, opened 2023-10-04
Decorator controller rewrite

misohu:
Signed commit from here (with discussion) https://github.com/canonical/kfp-operators/pull/342


i-chvets (approved):
Approved based on discussion in https://github.com/canonical/kfp-operators/pull/342

Issue #352, opened 2023-10-16
The recurring run does not work at kubeflow 1.17


emrahkyn:
### Bug Description

Hi Team,

I deployed a kubeflow 1.7 in test environment and was able to create a run without any issues. However, when I followed the similar procedure for recurring runs, it's created but never triggered to be executed. I could not get any error message, do you have any suggestions?
![image](https://github.com/canonical/kfp-operators/assets/76430580/4d2cd036-4f39-4e85-95b5-86d36d089830)

Thanks,
Emre

### To Reproduce

I deployed the kubeflow 1.7 by following the guideline [1] on kubernetes cluster and started to test it without doing any additional changes.

[1] https://charmed-kubeflow.io/docs/install

### Environment

It works on charmed kubernetes 1.28 with metallb load balancer.
[kfp-schedwf.log](https://github.com/canonical/kfp-operators/files/12915729/kfp-schedwf.log)
[kfp-api.log](https://github.com/canonical/kfp-operators/files/12915730/kfp-api.log)

### Relevant log output

```shell
The log files from kfp-api and kfp-schedwf pods are attached here. Please ignore the last part of the log files as I tried to create pipeline, experiments and run it manually using the kfp cli, but it did not work due to probably missing parameters.
```

### Additional context

I could not verify it yet but I've been informed that the recurring run does work at kubeflow 1.6 but after upgrade to 1.7, it started to not work.

DnPlas:
Hi @emrahkyn thanks for reporting this. I was actually able to reproduce this issue, it seems like https://github.com/canonical/kfp-operators/pull/329 introduced an issue with the kfp-schedwf workload image.

As a workaround while we fix it you can deploy this component with `juju deploy kfp-schedwf --channel 1.7/stable --resource oci-image=gcr.io/ml-pipeline/scheduledworkflow:2.0.0-alpha.7`

cc: @i-chvets

i-chvets:
Here are the findings:
- `kfp-schedwf` in 1.7 is PodSpec version.
- ROCKs were introduced (backported) to KFP pipelines for 1.7 #329 
- Container name in 1.7 `kfp-schedwf` charm differs from service name in ROCK:

Proposed solution:
- Change ROCK service name to match 1.7 container name. This will break `latest`, because service name != container name in sidecar version of `kfp-schedwf`. As a result, change service name to match container name in `main` OR create another version of ROCK for `main` (and 1.8).

emrahkyn:
[...]

PR #353, opened 2023-10-20
fix: bump kfp-schedwf rock version with patch

NohaIhab:
bump `kfp-schedwf` rock to v2 with the fix to issue https://github.com/canonical/pipelines-rocks/issues/38
partially addresses https://github.com/canonical/kfp-operators/issues/352


i-chvets (approved):
Approved.

PR #354, opened 2023-10-23
refactor: refactor env variables for kfp-api to match v2.0.2

DnPlas:
For a few releases the KFP API server component has been defining certain configurations using  a config file  instead of env variables. This change plus adding/removing vars will ensure the charmed kfp-api keeps consistent with the upstream project.

Please refer to the [upstream manifests](https://github.com/kubeflow/manifests/blob/master/apps/pipeline/upstream/base/pipeline/ml-pipeline-apiserver-deployment.yaml#L19) for more details on what should be configured.


charms/kfp-api/src/charm.py
```diff
@@ -210,64 +207,63 @@ def _kfp_api_layer(self) -> Layer:
...
+            "DBCONFIG_DBNAME": db_data["db_name"],
+            "DBCONFIG_HOST": db_data["db_host"],
+            "DBCONFIG_PORT": db_data["db_port"],
+            "DBCONFIG_CONMAXLIFETIME": "120s",  # 120s
+            "DB_DRIVER_NAME": "mysql",  # mysql
+            "DBCONFIG_MYSQLCONFIG_USER": db_data["db_username"],  # root
+            "DBCONFIG_MYSQLCONFIG_PASSWORD": db_data["db_password"],
+            "DBCONFIG_MYSQLCONFIG_DBNAME": db_data["db_name"],  # mlpipeline
+            "DBCONFIG_MYSQLCONFIG_HOST": db_data["db_host"],  # mysql
+            "DBCONFIG_MYSQLCONFIG_PORT": db_data["db_port"],  # 3306
```
orfeas-k, line 254: Regarding comments with expected values, do we need them here? Are they here by accident?

charms/kfp-api/src/charm.py
```diff
@@ -210,64 +207,63 @@ def _kfp_api_layer(self) -> Layer:
...
+            "OBJECTSTORECONFIG_SECRETACCESSKEY": object_storage["secret-key"],
+            "DEFAULTPIPELINERUNNERSERVICEACCOUNT": "default-editor",
             "MULTIUSER": "true",
-            "ML_PIPELINE_VISUALIZATIONSERVER_SERVICE_HOST": viz["service-name"],
-            "ML_PIPELINE_VISUALIZATIONSERVER_SERVICE_PORT": viz["service-port"],
+            "VISUALIZATIONSERVICE_NAME": viz_data["service-name"],
+            "VISUALIZATIONSERVICE_PORT": viz_data["service-port"],
+            "ML_PIPELINE_VISUALIZATIONSERVER_SERVICE_HOST": viz_data["service-name"],
+            "ML_PIPELINE_VISUALIZATIONSERVER_SERVICE_PORT": viz_data["service-port"],
+            "CACHE_IMAGE": self.model.config["cache-image"],
```
orfeas-k, line 263: I don't see those values in upstream deployment of `ml-pipeline` (I only see `CACHE_IMAGE` in `cache-server` deployment). Seeing though that you committed those afterwards, I understand that we may need them.

charms/kfp-api/tests/unit/test_operator.py
```diff
@@ -432,21 +434,17 @@ def test_install_with_all_inputs_and_pebble(
         )
         assert pebble_exec_command == f"bash -c '{exec_command}'"
         test_env = pebble_plan_info["services"][KFP_API_SERVICE_NAME]["environment"]
-        # there should be 1 environment variable
-        assert 1 == len(test_env)
         assert "test_model" == test_env["POD_NAMESPACE"]
```
orfeas-k, line 437: not sure if it would make sense to test more of the environment here

orfeas-k (commented):
Thank you @DnPlas, left some comments.

charms/kfp-api/src/charm.py
```diff
@@ -301,28 +297,6 @@ def _check_status(self):
                 raise ErrorWithStatus("Workload failed health check", MaintenanceStatus)
             self.model.unit.status = ActiveStatus()
 
-    def _upload_files_to_container(self, config_json):
-        """Upload required files to container."""
-        try:
-            self._check_container_connection(self.container)
```
orfeas-k, line 307: We should also remove this function if we don't need it anymore

charms/kfp-api/src/charm.py
```diff
@@ -210,64 +207,63 @@ def _kfp_api_layer(self) -> Layer:
...
-        """Generate configuration based on supplied data.
+    def _generate_environment(self) -> dict:
+        """Generate environment based on supplied data.
 
         Configuration is generated based on:
         - Supplied interfaces.
         - Database data: from MySQL relation data or from data platform library.
         - Model configuration.
+
+        Return:
```
orfeas-k, line 218: Good one, we should always try to have a proper docstring in our functions. :clap:

orfeas-k:
Also, I get that PR description is inaccurate
> has been defining certain configurations using env variables instead of a config file.

charms/kfp-api/src/charm.py
DnPlas: I left them behind as a reference, do you think they are misleading?

charms/kfp-api/tests/unit/test_operator.py
DnPlas: I was also wondering how and if it's really needed. The env rn may be different in the next release, so I did not bother checking that the number of variables were X. If you have suggestions on how we can test this better, let me know.

charms/kfp-api/src/charm.py
DnPlas: Agreed.

charms/kfp-api/src/charm.py
orfeas-k: Ok, seems like sth could be done by `cache-server` deployment which we do not deploy if I 'm not mistaken

orfeas-k: (you may resolve this once you see this)

charms/kfp-api/tests/unit/test_operator.py
DnPlas: I have added an assertion in the test case that checks the pebble layer. Kindly review it ^_^

orfeas-k (approved):
Great work, approved!

PR #358, opened 2023-10-25
fix: add correct SA name to CRB

DnPlas:
The previous CRB was pointing to a SA that is not used by the api server for authentication. The rename was introduced by a244ffe.


misohu (approved):
LGTM .... maybe in future we can add a test facing this

PR #364, opened 2023-10-27
chore: updates the test bundle definitions and use latest/edge for tests

DnPlas:
This commit updates the many bundle defintions in tests/integration/bundles/ to match the latest changes to the repository.
At the same time it changes the bundle that is used for running repository level integration tests from 1.7/stable to latest/edge to be able to test the kfp-operators and all other related charms with the latest changes.


tests/integration/bundles/kfp_1.8_stable_install.yaml.j2
ca-scribner, line 1: Because we've changed what `channel: 2.0/stable` means (it used to be 1.7's kfp release, but now it is 1.8's kfp release), *at time of approving this PR* this bundle is actually the KFP from KF 1.7, **not** KF 1.8.  But in a week or two once we cut our Kubeflow 1.8 release, `channel: 2.0/stable` will retroactively be populated with the KFP release from KF 1.8.  

I don't know if there's anything we can (or should) do about this here, but raising it here because it feels awkward and it should at least be documented somewhere.

tests/integration/bundles/kfp_latest_edge.yaml.j2
```diff
@@ -1,13 +1,35 @@
...
+    trust: true
+  kubeflow-roles:
+    charm: kubeflow-roles
+    channel: latest/edge
+    scale: 1
+    trust: true
 {%- if local_build == false %}
   kfp-api:                 { charm: ch:kfp-api, channel: latest/edge, scale: 1, trust: true}
-  kfp-profile-controller:  { charm: ch:kfp-profile-controller, channel: latest/edge, scale: 1, trust: true }
+  kfp-metadata-writer:     { charm: ch:kfp-metadata-writer, channel: latest/edge, scale: 1, trust: true}
```
ca-scribner, line 32: I know we debated this a [a few months back](https://github.com/canonical/kfp-operators/pull/238#issuecomment-1624037301), but I'd still vote to keep `kfp-profile-controller` in here.  I don't think kfp functions completely without it, even if we don't have a test yet that spots the parts that are broken.  I prefer keeping our test env close to our prod env, even if we can't point to a specific reason for it.

If we do remove kfp-profile-controller from this bundle as proposed, then I think we should also remove `metacontroller-operator`.  `metacontroller-operator` is included only because it is a dependency of `kfp-profile-controller`

ca-scribner, line 32: Apart from seeing kfp come alive, do we have any tests that prove the metadata writer is working correctly?  Maybe running a pipeline and then exporting some metadata to prove it was logged correctly?  Or a manual test that shows in the UI where to see the data?

ca-scribner (approved):
generally lgtm, but added a few comments.  Main comment is that metadata component should probably have a test to demonstrate it is working (ideally an automated test, but could also be manual)

tests/integration/bundles/kfp_1.8_stable_install.yaml.j2
DnPlas: Are you suggesting that we should just wait until we release to add this bundle? We can do that, I understand the reasons why we'd want that.

tests/integration/bundles/kfp_latest_edge.yaml.j2
DnPlas: I can add it back. We can increase our test coverage later if we want.

DnPlas: We don't have any, but we can definitely add one or two test cases for the metadata-writer, though that seems to belong in a different PR.

DnPlas:
> generally lgtm, but added a few comments. Main comment is that metadata component should probably have a test to demonstrate it is working (ideally an automated test, but could also be manual)
> 
> This PR is also a little awkward to approve because `Test the bundle` fails, so I can't easily see the integration tests are still healthy.

The tests on this PR should pass, I will verify why that is not the case, but thanks for the review!

tests/integration/bundles/kfp_1.8_stable_install.yaml.j2
ca-scribner: Honestly I'm not sure what is best.  Strictly speaking waiting is maybe better, but practically speaking I don't think anyone will use this bundle until after we release anyway so it wont matter.  I'm good either way as long as we know what's happening

tests/integration/bundles/kfp_latest_edge.yaml.j2
ca-scribner: I feel better with kfp-profile-controller in the test, but trust your judgement so whichever you choose is fine with me.  

But if you choose to remove kfp-profile-controller, then we should also remove metacontroller-operator

ca-scribner: Agreed.  added as #384 and resolving this

DnPlas:
Merging despite the errors in the integration tests. We have identified why they are failing, but the fix will land in #383 .

Issue #367, opened 2023-11-01
kfp-api's `apiserver` initially fails to connect to minio service during bundle deploy


ca-scribner:
### Bug Description

When deploying a bundle including kfp, kfp-api's `apiserver` initially fails to connect to the minio service with the following error in the container's logs from pebble:

```
2023-10-28T13:51:01.193Z [apiserver] F1028 13:51:01.193216     131 minio.go:76] Failed to create Minio client. Error: Error while creating minio client: Endpoint:  does not follow ip address or domain name standards.: Endpoint:  does not follow ip address or domain name standards.
```

This is repeated for some period of time (~1hr in the provided log during full kubeflow deployment) and eventually resolves itself

### To Reproduce

Deploy kfp-api with the following bundle:

Notes:
* probably we can shrink this down to just kfp and its dependencies, although if there's a race condition it might not show up the same way without the load of deploying everything else)
* kfp charms here came from an ephemeral edge branch `latest/edge/pr-361`, which is roughly equivalent to [this code](https://github.com/canonical/kfp-operators/tree/1.8-updates-dev-branch)

<details>
  <summary>bundle.yaml</summary>

```
bundle: kubernetes
applications:
  admission-webhook:
    charm: admission-webhook
    channel: latest/edge
    revision: 269
    series: focal
    resources:
      oci-image: 272
    scale: 1
    constraints: arch=amd64
    trust: true
  argo-controller:
    charm: argo-controller
    channel: latest/edge
    revision: 393
    series: focal
    resources:
      oci-image: 343
    scale: 1
    constraints: arch=amd64
    trust: true
  dex-auth:
    charm: dex-auth
    channel: latest/edge
    revision: 353
    series: focal
    resources:
      oci-image: 447
    scale: 1
    options:
      public-url: http://10.64.140.43.nip.io
      static-password: admin
      static-username: admin
    constraints: arch=amd64
    trust: true
  envoy:
    charm: envoy
    channel: latest/edge
    revision: 91
    series: focal
    resources:
      oci-image: 91
    scale: 1
    constraints: arch=amd64
  istio-ingressgateway:
    charm: istio-gateway
    channel: latest/edge
    revision: 715
    series: focal
    scale: 1
    options:
      kind: ingress
    constraints: arch=amd64
    trust: true
  istio-pilot:
    charm: istio-pilot
    channel: latest/edge
    revision: 704
    series: focal
    scale: 1
    options:
      default-gateway: kubeflow-gateway
[...]

PR #370, opened 2023-11-01
Update bug_report.yml to remove shell attribute in logs

ca-scribner:
Logs are saved as shell input, which means they get rendered directly as a code block in the final output.  Let's change this to not be rendered.

The motivation for this is to enable something like [this issue's logs](https://github.com/canonical/kfp-operators/issues/367#issue-1972405752), where we can use a `<details> ... </details>` block to hide the logs in an expandable text block.  If the logs are rendered as shell, the user cannot do this.


DnPlas:
Thanks @ca-scribner , this change does not apply anymore. We are now using a different template. Closing.

PR #378, opened 2023-11-03
reformat kfp-profile-controller's sync.py to match upstream

ca-scribner:
kfp-profile-controller implements a sync.py file from upstream which we lightly modify.  This commit reformats that sync.py to be as similar as possible to upstream's sync.py for easier comparison and maintenance.  #331 previously reformatted sync.py to meet our style guide, but that made it harder to track and changes upstream and import them into our version

This should have no functional change on the charm, it only changes things to make maintenance easier.


DnPlas (approved):
Thanks @ca-scribner, LGTM!

PR #380, opened 2023-11-08
todo

DnPlas:
todo


charms/kfp-api/src/templates/ml-pipeline-service.yaml.j2
ca-scribner, line 1: This PR should work as described, but as I review I personally feel the change is unnecessary and could lead to a future developer making easy and subtle mistakes.  I would close this PR and leave the old method in place.  

My main concern is that we typically can rely on the charm's service having the same name as the deployed application by convention in juju, but wont be able to here.  We see that relaince [in kfp-api](https://github.com/canonical/kfp-operators/blob/3b3da165b60a011ce6826bd98133d540c0897a28/charms/kfp-api/src/charm.py#L60) and in other charms like [seldon](https://github.com/canonical/seldon-core-operator/blob/7b01622a82dbc40cca9242aac02794a4775388e7/src/charm.py#L126).  Not saying that we cannot rename the service like done in this PR, but that I can see someone making a mistake in the future because they expect this charm to behave like our other charms.

While I agree that the previous way deployed one more `Service` than absolutely necessary, I prefer it over this PR because the charm behaves like our other charms.  The previous way worked for essentially no cost (having an extra `Service` is not expensive) and did not introduce a special case where kfp-api has a different service name convention than everything else.

ca-scribner (changes requested):
Functionally lgtm, but a few minor cleanups to do.  

Also added a comment about how I personally would skip this PR at all, but if this is what we want to do then lgtm

charms/kfp-api/src/charm.py
```diff
@@ -38,9 +38,9 @@
 METRICS_PATH = "/metrics"
 PROBE_PATH = "/apis/v1beta1/healthz"
 
+APISERVER_SVC_NAME = "ml-pipeline"
```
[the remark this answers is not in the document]
DnPlas: +1

charms/kfp-api/src/templates/ml-pipeline-service.yaml.j2
DnPlas: > My main concern is that we typically can rely on the charm's service having the same name as the deployed application by convention in juju, but wont be able to here. 

While that is true for some charms, some components like the kfp-operators expect the service name of the `apiserver` to be named `ml-pipeline` (see https://github.com/canonical/kfp-operators/pull/237 for more information), which already changes the "pattern" that we have in other places. 

> Not saying that we cannot rename the service like done in this PR, but that I can see someone making a mistake in the future because they expect this charm to behave like our other charms.

I actually see the opposite, as there could be charms that get a Service named after the application, but in reality the components that talk to that application require a different name. Trying in-cluster communication to the pipelines `apiserver` is a good example of this.

> While I agree that the previous way deployed one more Service than absolutely necessary, I prefer it over this PR because the charm behaves like our other charms. The previous way worked for essentially no cost (having an extra Service is not expensive) and did not introduce a special case where kfp-api has a different service name convention than everything else.

I agree that it is not expensive to have two Services, but at the same time having an extra manifest to render and apply can introduce errors (as we have seen before) in places like the selector (see [here](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/src/templates/ml-pipeline-service.yaml.j2#L17) it is incorrect) and the name (as we have discussed before). It also creates an extra resource that we have to take care of (maintaining, updating), while a juju created Service is there by default.

kimwnasptd: Personally I'd like to reduce any manifest yamls we apply as much as possible, especially for K8s Services where we've been bitten a lot of times by using incorrect label selectors. Plus it seems weird to me to have 2 K8s Services for the same Pods.

I see @ca-scribner's point though on how now we will be deviating from juju's standard naming scheme. But at the same time, with my current understanding, all Charm<>Juju communications happen from Charm -> Juju Controller. So for JuJu Controller there's no need to know/use the Charm's K8s Service. But at the same time maybe I'm missing something on why each Charm then gets a Service by default.

Only edge case I would consider here is COS, in which another Pod needs to reach the Charm workloads. So what I propose is:
1. If COS doesn't use K8s Services for getting metrics
2. If Juju doesn't use the K8s Service at all

Then let's continue with this PR since it will reduce the K8s resources we apply

kimwnasptd: But one more small question here. When we do `juju relate` then how are the two Charm pods communicating with each other? Is there a communication between the two pods? And if yes, do they use the K8s service or direct Pod IPs to talk with each other?

ca-scribner: for `juju relate` we should check, but I don't think juju makes two charm pods ever communicate directly.  CharmA talks to CharmB by passing data to a relation/pulling data from it.  Everything is initiated from the Charm so for that, we should be ok (we've discussed this same issue when talking about restricting traffic via istio)

ca-scribner: To me, this PR feels like we are adding the risk of an unknown unknown.  We are taking an object managed by some other application (`svc/application-name` managed by Juju) and deleting it, then adding our own of a different name).  From a first principles standpoint this should raise a warning flag because we are changing an object we don't control.  That doesn't mean we **can't** do this, but it means imo we should have a compelling reason to do it and I don't see our reason as compelling enough.

There's a fair chance this PR works as expected, but since we're deleting stuff we don't own there's also some chance that it breaks in a way we don't know.  As a Juju dev, why shouldn't they in future think "we create a service called `svc/application-name`, lets refactor how our networking model works and send traffic to that service"?

DnPlas:
Based on an extensive investigation with the Juju and Observability teams, we have concluded the following:

1. In terms of Observability, the Service name is not assumed to always be the same as the application, therefore, changing it would not be an issue.
2. In terms of Juju (based on a [public discussion with the team](https://chat.charmhub.io/charmhub/pl/b44rorgwfj8hurajyfehc3kgoa)), renaming the `Service` won't likely be a problem to Juju, but it is discouraged. Juju will not point at the `Service`, but to keep consistency, Juju names everything with the same name at deployment time. 

Another important note is that `juju` will "re-create" this `Service` on event hooks that are not necessarily observed by the `KubernetesServicePatch`, which can cause downtimes or complete disconnections when other applications try to reach the "patched" `Service`.

Overall we can conclude that, while it won't mean a huge issue with juju or Observability applications, it is best to create services if they need to have a determined form. We can close this PR based on that conclusions, but we could continue the discussion about how the `KubernetesServicePatch` library is not providing a guarantee that it will work always, so we have to revisit our charms that use this library.

PR #386, opened 2023-11-17
fix: reduce flakiness of kfp-api integration tests

ca-scribner:
The mysql-k8s charm during deployment sometimes lands temporarily in an error state.  This error state resolves over time and the charm continues on to work as expected, but causes our integration tests to fail (like [this example](https://github.com/canonical/kfp-operators/actions/runs/6907962624/job/18796234535?pr=385#step:5:232)) because we do `wait_for_idle(['mysql-k8s'], ...raise_on_error=True)`.  ~~This fix sets `raise_on_error=False` to avoid this.~~

~~This does not change the passing condition of the test - the test will only pass if mysql-k8s goes to Active/Idle.  The only change is that mysql-k8s going to Error state does not result in a fast failure - the test will continue until the timeout before failing.~~

Edit on 2023-11-20: rather than set `raise_on_error=True`, the solution is now to bump the mysql version to `latest/stable` as it did not show the same same flaky behaviour.

### DIFF
--- a/charms/kfp-api/tests/integration/test_charm.py
+++ b/charms/kfp-api/tests/integration/test_charm.py
@@ -86,9 +86,6 @@ class TestCharm:
         await ops_test.model.wait_for_idle(
             apps=["mysql-k8s"],
             status="active",
-            # mysql-k8s sometimes goes into error and then resolves itself.  Do not raise on error
-            # here so we don't flake the test when the situation will resolve itself.
-            raise_on_error=False,
             raise_on_blocked=True,
             timeout=90 * 30,
             idle_period=20,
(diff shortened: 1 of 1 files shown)


misohu:
This makes me concerned why is it going to error state? Should not we open issue on mysql charm? No problem with the change just to notify mysql team about it.

ca-scribner:
100% agree @misohu - but I forgot to link it here!  https://github.com/canonical/mysql-k8s-operator/issues/337

DnPlas:
Hey @ca-scribner I just changed the channel we are using for deploying `mysql-k8s` to be on sync with what we know works. This should be a better solution for this issue.

DnPlas (approved):

ca-scribner:
During review, @dnplas found that the flakiness was solved by bumping `mysql-k8s` to `latest/stable`, so the PR has been updated to this.

PR #385, opened 2023-11-17
refactor kfp-api's on_remove to use KRH built-in delete

ca-scribner:
Refactors kfp-api charm's `on_remove()` event handler to use KubernetesResourceHandler.delete() rather than its own removal procedure.  This was done to:
* standardise on a removal procedure with other charms
* remove the need for the `on_remove()` handler to have a KRH object with full `context`.  In future we will add another Kubernetes object here that requires relation data to render, and this relation data may not be available during the removal event

PR #387, opened 2023-11-17
add service `minio-service` to kfp api to patch upstream kfp bug

ca-scribner:
This adds to kfp-api a service called `minio-service` which points to the related object-storage's s3 service.  This has been added to address a bug in upstream kfp, as explained [here](https://github.com/canonical/minio-operator/pull/151).  This service was originally added to the minio charm in [minio pr 151](https://github.com/canonical/minio-operator/pull/151), but has been refactored so it is added here instead as described in [minio issue 153](https://github.com/canonical/minio-operator/issues/153).

### Reviewer notes
* All lint and unit tests should pass.  The kfp-api integration tests should pass, but the bundle tests may fail (are being fixed in a separate parallel PR)
* to manually test:
  * deploy kfp-api and required dependencies (an easy way to do this is to run `tox -e integration -- --model kubeflow -k "build_and_deploy or relational_db_relation_with_mysql_relation"`, or just to manually do the same deployments as the integration test)
  * see if the `svc/minio-service` exists (`kubectl get svc minio-service`)
  * manually use minio through the service
    * deploy a pod with the minio client (mc) and run bash in the pod: `kubectl run mc -it --rm --image minio/mc --command -- /bin/bash`
    * in the pod, do: 
      * `mc alias set myMinio http://minio-service.kubeflow:9000 ACCESSKEY SECRETKEY` (get accesskey/secretkey from juju config by inspecting kfp-api's relation with the minio charm)
      * `mc mb myMinio/testbucket`
      * `mc ls myMinio`  # (should see your testBucket)

~~Merge after #385 and #386.~~ - Done

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -100,7 +100,10 @@ jobs:
           charmcraft-channel: latest/candidate
 
       - run: |
-          sg snap_microk8s -c "tox -e ${{ matrix.charm }}-integration"
+          # Requires the model to be called kubeflow due to
+          # https://github.com/canonical/kfp-operators/issues/389
+          juju add-model kubeflow
+          sg snap_microk8s -c "tox -e ${{ matrix.charm }}-integration -- --model kubeflow"
 
       - name: Collect charm debug artifacts
         uses: canonical/kubeflow-ci/actions/dump-charm-debug-artifacts@main
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -113,6 +113,7 @@ jobs:
     name: Test the bundle
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
       matrix:
         sdk:
           - v1
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -99,7 +99,8 @@ jobs:
           juju-channel: 3.1/stable
           charmcraft-channel: latest/candidate
 
-      - run: |
+      - name: Integration tests
+        run: |
           # Requires the model to be called kubeflow due to
           # https://github.com/canonical/kfp-operators/issues/389
           juju add-model kubeflow
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -77,14 +77,6 @@ class KfpApiOperator(CharmBase):
         self._database_name = "mlpipeline"
         self._container = self.unit.get_container(self._container_name)
 
-        # setup context to be used for updating K8S resources
-        self._context = {
-            "app_name": self._name,
-            "namespace": self._namespace,
-            "service": self._name,
-            "grpc_port": self._grcp_port,
-            "http_port": self._http_port,
-        }
         self._k8s_resource_handler = None
 
         grpc_port = ServicePort(int(self._grcp_port), name="grpc-port")
@@ -159,6 +151,21 @@ class KfpApiOperator(CharmBase):
     def container(self):
         """Return container."""
         return self._container
+
+    @property
+    def _context(self):
+        """Return the context used for generating kubernetes resources."""
+        interfaces = self._get_interfaces()
+        os = self._get_object_storage(interfaces)
+
+        context = {
+            "app_name": self._name,
+            "namespace": self._namespace,
+            "service": self._name,
+            "grpc_port": self._grcp_port,
+            "http_port": self._http_port,
+        }
+        return context
 
     @property
     def k8s_resource_handler(self):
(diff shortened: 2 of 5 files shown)


charms/kfp-api/src/charm.py
```diff
@@ -142,11 +139,38 @@ def __init__(self, *args):
...
+        os = self._get_object_storage(interfaces)
+
+        context = {
+            "app_name": self._name,
+            "namespace": self._namespace,
+            "service": self._name,
+            "grpc_port": self._grcp_port,
+            "http_port": self._http_port,
+            # Must include .svc.cluster.local for DNS resolution
+            "minio_url": f"{os['service']}.{os['namespace']}.svc.cluster.local",
```
DnPlas, line 169: The hardcoded value in upstream expects the Namespace value to always be `kubeflow` for the minio Service. I wonder if we should add a check in the kfp-api charm code to ensure the object store shares this expected value, otherwise we can block the unit. I wouldn't expect this to happen often as these two charms should be deployed in the same model, but I wonder if we should ensure this on the "requirer" side.

misohu:
So this PR https://github.com/canonical/minio-operator/pull/151 creates a service with lightkube to create svc, leading to duplicate services `minio` and `minio-service` (check the discussion). The orginal PR could not use the `KubernetesServicePatcher` as it was a podspec and lightkube needs `trust`. Your PR @ca-scribner is using sidecar charm so we can use `trust`. I think we should use the `KubernetesServicePatcher` as we do in other charms.

DnPlas (commented):
Thanks @ca-scribner, this makes sense. Just a tiny comment.

charms/kfp-api/tests/integration/test_charm.py
```diff
@@ -86,6 +86,9 @@ async def test_relational_db_relation_with_mysql_relation(self, ops_test: OpsTes
         await ops_test.model.wait_for_idle(
             apps=["mysql-k8s"],
             status="active",
+            # mysql-k8s sometimes goes into error and then resolves itself.  Do not raise on error
```
DnPlas, line 89: This change is already covered in #386, we can track this there.

charms/kfp-api/src/charm.py
```diff
@@ -554,14 +585,20 @@ def _on_upgrade(self, _):
...
+        k8s_resource_handler = KubernetesResourceHandler(
+            field_manager=self._lightkube_field_manager,
+            template_files=[],
+            context={},
+            logger=self.logger,
+            labels=self._charm_default_kubernetes_labels,
+            resource_types={Service, ServiceAccount, ClusterRole, ClusterRoleBinding},
+        )
+        load_in_cluster_generic_resources(k8s_resource_handler.lightkube_client)
+        k8s_resource_handler.delete()
```
DnPlas, line 600: This change is covered in #385, we should track this change there

charms/kfp-api/src/charm.py
ca-scribner: Good point.  What do you think about adding [this check](https://github.com/canonical/kubeflow-dashboard-operator/blob/a6f9dc0968ded793a1692e10766e4454f4c9127d/src/charm.py#L184) to kfp-api, and then hard coding this to `kubeflow` and adding a note?

charms/kfp-api/tests/integration/test_charm.py
ca-scribner: Yup I'll remove now that #386 went another direction

charms/kfp-api/src/charm.py
ca-scribner: This change is tracked in #385.  This PR is based on #385.  When #385 is merged, this should no longer show up as a change.

charms/kfp-api/src/charm.py
```diff
@@ -142,11 +139,38 @@ def __init__(self, *args):
...
     @property
     def container(self):
         """Return container."""
         return self._container
 
+    @property
+    def _context(self):
+        """Return the context used for generating kubernetes resources."""
+        interfaces = self._get_interfaces()
+        os = self._get_object_storage(interfaces)
```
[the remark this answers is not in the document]
ca-scribner: I think this is a comment for #385 rather than here.  Adding a note there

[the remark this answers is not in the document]
ca-scribner: actually I got confused, it is not in #385.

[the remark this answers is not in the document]
ca-scribner: Done

charms/kfp-api/src/charm.py
ca-scribner: I think it is clearer now that things have been rebased.

charms/kfp-api/src/charm.py
ca-scribner: Added the check - let me know what you think

.github/workflows/integrate.yaml
```diff
@@ -100,7 +100,10 @@ jobs:
           charmcraft-channel: latest/candidate
 
       - run: |
```
NohaIhab, line 102: nit: let's add a name to this step so that the comment doesn't show up as the title of the step in the CI

NohaIhab: I put in a fix for this

orfeas-k (approved):

PR #393, opened 2023-11-27
Add badges to the repo

kimwnasptd:
This will ensure that we comply with the expectations in https://releases.juju.is/?team=MLOps


misohu (commented):
Hey @kimwnasptd I think you missed the kfp-metadata-writer

misohu (approved):

PR #415, opened 2024-03-29
self-hosted PoC

kimwnasptd:
Create a super simple example of why self-hosted runners don't work with
- microK8s
- lightkube client

Aim only for debug and reference purposes


DnPlas:
@kimwnasptd are you still working on this PR? If this is for testing purposes only, please convert it to draft or close it if it is nor required anymore.

Issue #407, opened 2024-02-06
Update the `k8s-service` relation interface to v2 in `kfp-metadata-writer`


DnPlas:
### Context

Based on the proposal in https://github.com/canonical/mlops-libs/issues/2 and the changes that are going to be introduced in https://github.com/canonical/mlmd-operator/issues/71 we'll need to update the `k8s-service` relation interface handling in the ``kfp-metadata-writer` charm.

### What needs to get done

Replace the SDI component that handles this relation interface with the library that is going to be created in https://github.com/canonical/mlops-libs/issues/2

### Definition of Done

1. The SDI implementation is replaced with the library proposed in https://github.com/canonical/mlops-libs/issues/2
2. There is a working upgrade path from a previous version of this charm

PR #435, opened 2024-04-17
chore: trust mlmd due to refactor

DnPlas:
mlmd was recently refactored to use the sidecar pattern, which now requires this charm to be trusted as it applies k8s resources.


DnPlas:
Closing this as we'll also need to refactor kfp-metadata-writer. Related to #407 .

PR #455, opened 2024-05-09
feat(api-server): Enable charm configuration of launcher and driver images

### LINKED ISSUE #452: Expose configuration of kfp-api's launcher and driver images in charm config in `main`
### Context

https://github.com/kubeflow/pipelines/pull/10269 enables administrators to configure the images used for kfp's launcher and driver containers - we need to expose that configuration as charm config so that users can use their own custom images.  This is important mainly for airgapped deployments, where these will be pulled from a local container registry.  

The UX of this is up to the developer.  These could be exposed either each as their own config, or as a dict like in [istio-pilot](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/config.yaml#L25) - whichever feels more natural for the UX.  The charm should default to the current upstream images. 

In completing this task, we must also update `tools/get-images.sh` to include these images, that way when doing routine scanning or building an airgapped tarball these images are included.  

### What needs to get done

.

### Definition of Done

1. config for launcher and driver images is exposed
2. get-images.sh script is updated to include these images
3. charm is in `main`

orfeas-k:
Enable charm configuration of launcher and driver images used during pipeline steps. Those can be configured thourgh values `launcher-image` and `driver-image`. When unset, KFP uses the default images defined in upstream code. This is based on upstream implementation introduced in kubeflow/pipelines#10269.

Closes #452

#### Testing
This was tested with upstream image `gcr.io/ml-pipeline/api-server:2.0.5`. The rest of kfp used the ROCKs.

It can be tested that pipeline steps are using the proper images by configuring the charm and then running a pipeline. 
```bash
juju config kfp-api launcher-image="fake-image"
juju config kfp-api driver-image="fake-image"
```

You should see pods created by the pipeline run that spin up containers using the images provided in the config options.

Note that:
* Each configuration should be tested separately. A pod using the `launcher-image` is only spun up after (at least some) `driver-image` pods are completed. Thus, `launcher-image` configuration can only be tested when `driver-image` config is healthy.
* When rerunning the same exact pipeline, KFP will use the launcher spun up previously, if one was successfully created, instead of spinning a new one (thus the previously used `launcher-image`).

### DIFF
--- a/charms/kfp-api/config.yaml
+++ b/charms/kfp-api/config.yaml
@@ -51,9 +51,11 @@ options:
     description: Default name of object storage bucket.
   launcher-image:
     type: string
-    default: ""
+    # Source: https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/v2/compiler/argocompiler/container.go#L27
+    default: "gcr.io/ml-pipeline/kfp-launcher@sha256:80cf120abd125db84fa547640fd6386c4b2a26936e0c2b04a7d3634991a850a4"
     description: Launcher image used during a pipeline's steps.
   driver-image:
     type: string
-    default: ""
+    # Source: https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/v2/compiler/argocompiler/container.go#L29
+    default: "gcr.io/ml-pipeline/kfp-driver@sha256:8e60086b04d92b657898a310ca9757631d58547e76bbbb8bfc376d654bef1707"
     description: Driver image used during a pipeline's steps.
--- a/charms/kfp-api/config.yaml
+++ b/charms/kfp-api/config.yaml
@@ -49,3 +49,11 @@ options:
     type: string
     default: "mlpipeline"
     description: Default name of object storage bucket.
+  launcher-image:
+    type: string
+    default: ""
+    description: Launcher image used during a pipeline's steps.
+  driver-image:
+    type: string
+    default: ""
+    description: Driver image used during a pipeline's steps.
--- a/charms/kfp-api/src/charm.py
+++ b/charms/kfp-api/src/charm.py
@@ -290,6 +290,8 @@ class KfpApiOperator(CharmBase):
             "OBJECTSTORECONFIG_HOST": f"{object_storage['service']}.{object_storage['namespace']}",
             "OBJECTSTORECONFIG_PORT": str(object_storage["port"]),
             "OBJECTSTORECONFIG_REGION": "",
+            "V2_LAUNCHER_IMAGE": self.model.config["launcher-image"],
+            "V2_DRIVER_IMAGE": self.model.config["driver-image"],
         }
 
         return env_vars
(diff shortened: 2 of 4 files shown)


charms/kfp-api/config.yaml
ca-scribner, line 1: We need to add the defaults for these images to the `tools/get-images.sh` script.  That script returns all images used in this repo (for scanning or airgapping purposes), but it doesn't know about these images natively

I see [istio](https://github.com/canonical/istio-operators/blob/main/tools/get-images.sh) setting the default value for the config here, then pulling that value in get-images.sh.  that is an option

charms/kfp-api/src/charm.py
```diff
@@ -290,6 +290,8 @@ def _generate_environment(self) -> dict:
             "OBJECTSTORECONFIG_HOST": f"{object_storage['service']}.{object_storage['namespace']}",
             "OBJECTSTORECONFIG_PORT": str(object_storage["port"]),
             "OBJECTSTORECONFIG_REGION": "",
+            "V2_LAUNCHER_IMAGE": self.model.config["launcher-image"],
+            "V2_DRIVER_IMAGE": self.model.config["driver-image"],
```
ca-scribner, line 294: I see our defaults atm are `""`, so we're setting `V2_*_IMAGE=""` for the normal case.  Am I understanding this correctly that the upstream code interprets that as us wanting their default image?  I think it must if its passing tests.

ca-scribner (dismissed):
changes lgtm but we need to update the `tools/get-images.sh` file as commented

NohaIhab:
I synced with @orfeas-k to test this, and it was working correctly for me!
We agreed to add to the PR the instructions that we tested with.

charms/kfp-api/src/charm.py
orfeas-k: Yep, confirming this part was what took me most of the time implementing this yesterday. Upstream uses default image since it interprets empty string as a not defined environment variable.

orfeas-k: In any case, we 'll update the default value of the config so this shouldn't be an issue.

charms/kfp-api/config.yaml
orfeas-k: As discussed live, I pushed a commit that includes the default value in the `config.yaml` file and updates the get-images.sh script which results in the following output.

```diff
charmedkubeflow/frontend:2.0.3-d4ac42b
charmedkubeflow/scheduledworkflow:2.0.3-7d6d3e4
charmedkubeflow/visualization-server:2.0.3-8169d0c
charmedkubeflow/api-server:2.0.3-e037d33
charmedkubeflow/viewer-crd-controller:2.0.3-d89d9fc
charmedkubeflow/persistenceagent:2.0.3-a3714a9
charmedkubeflow/metadata-writer:2.0.3-286d4a8
python:3.7
gcr.io/google-containers/busybox
+ gcr.io/ml-pipeline/kfp-launcher@sha256:80cf120abd125db84fa547640fd6386c4b2a26936e0c2b04a7d3634991a850a4
+ gcr.io/ml-pipeline/kfp-driver@sha256:8e60086b04d92b657898a310ca9757631d58547e76bbbb8bfc376d654bef1707
```

kimwnasptd:
Deployed as well the Charm and managed to change the images successfully.

kimwnasptd (approved):

PR #467, opened 2024-05-22
backport(feat): Enable configuration in kfp-api of launcher and driver images from #455

### LINKED ISSUE #452: Expose configuration of kfp-api's launcher and driver images in charm config in `main`

### LINKED ISSUE #453: Expose configuration of kfp-api's launcher and driver images in charm config in `track/2.0`
### Context

backport #452 to `track/2.0`

### What needs to get done

.

### Definition of Done

1. #452 is backported to `track/2.0`

orfeas-k:
Enable charm configuration of launcher and driver images used during pipeline steps. Those can be configured through config options `launcher-image` and `driver-image`. When unset, KFP uses the default values defined. This is based on upstream implementation introduced in kubeflow/pipelines#10269.

Closes #453

### DIFF
--- a/charms/kfp-api/config.yaml
+++ b/charms/kfp-api/config.yaml
@@ -49,3 +49,13 @@ options:
     type: string
     default: "mlpipeline"
     description: Default name of object storage bucket.
+  launcher-image:
+    type: string
+    # Source: https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/v2/compiler/argocompiler/container.go#L27
+    default: "gcr.io/ml-pipeline/kfp-launcher@sha256:80cf120abd125db84fa547640fd6386c4b2a26936e0c2b04a7d3634991a850a4"
+    description: Launcher image used during a pipeline's steps.
+  driver-image:
+    type: string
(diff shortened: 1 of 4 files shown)


NohaIhab (approved):
tested in #455 and will be tested further in airgapped as part of https://github.com/canonical/bundle-kubeflow/issues/889

PR #478, opened 2024-05-24
Add repo-level CONTRIBUTING.md file including information for Updating manifests

orfeas-k:
Note that there are CONTRIBUTING.md files in each of the charms, but since changes usually take place at the same time for all of the charms, I added this file at the repo-level.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,35 @@
+## Update manifests
+
+KFP charms use Jinja2 templates in order to store manifests that need to be applied during its deployment. The process for updating them is:
+
+### Spot the differences between versions of a manifest file
+
+1. Install `kustomize` using the official documentation [instructions](https://kubectl.docs.kubernetes.io/installation/kustomize/)
+2. Clone [Kubeflow manifests](https://github.com/kubeflow/manifests) repo locally
+3. `cd` into the repo and checkout to the branch or tag of the target version.
+4. Build the manifests with `kustomize` and save the file:
+
+```bash
+kustomize build apps/pipeline/upstream/env/cert-manager/platform-agnostic-multi-user > kfp-manifests-vX.yaml
+```
+
+5. Checkout to the branch or tag of the version of the current manifest
+6. Build the manifest with `kustomize` (see step 4) and save the file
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+kustomize build apps/pipeline/upstream/env/cert-manager/platform-agnostic-multi-user > kfp-manifests-vX.yaml
+```
+
+5. Checkout to the branch or tag of the version of the current manifest
+6. Build the manifest with `kustomize` (see step 4) and save the file
+7. Compare both files to spot the differences (e.g. using diff `diff kfp-manifests-vX.yaml kfp-manifests-vY.yaml > kfp-vX-vY.diff`)
+
+### Introduce changes
+
+Once the comparison is done, add any changes to the relevant ClusterRoles and ClusterRoleBindings to the
```
kimwnasptd, line 22: @orfeas-k can we add a comment here and mention that we do this only for the aggregated ClusterRoles and ClusterRoleBindings? 

We want to ensure users won't copy ClusterRoles that upstream had for the ServiceAccount of the Deployment (since in our case JuJu creates a SA for the sidecar Charm)

DnPlas: Agreed.

CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Spot the differences between versions of a manifest file
+
+1. Install `kustomize` using the official documentation [instructions](https://kubectl.docs.kubernetes.io/installation/kustomize/)
+2. Clone [Kubeflow manifests](https://github.com/kubeflow/manifests) repo locally
+3. `cd` into the repo and checkout to the branch or tag of the target version.
+4. Build the manifests with `kustomize` and save the file:
+
+```bash
+kustomize build apps/pipeline/upstream/env/cert-manager/platform-agnostic-multi-user > kfp-manifests-vX.yaml
```
DnPlas, line 13: Since this may change in the future, would it make sense to add a link to the pipelines section of the readme of `kubeflow/manifests` repo to show what we take as reference?

DnPlas (commented):
Few nits, other than that LGTM! Thanks @orfeas-k !

DnPlas (approved):
Thanks @orfeas-k !

PR #486, opened 2024-05-31
fix: set correct default for custom_images in kfp-profile-controller

### LINKED ISSUE #484: `kfp-profile-controller` `custom_images` config default is incorrect for frontend image
### Bug Description

the default of the `custom_images` config for `kfp-profile-controller` charm is:
```
    default: | 
      visualization_server : ''
      frontend_image : ''
```
Meanwhile, the actual key used in the [json defining the default images](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-profile-controller/src/default-custom-images.json) is:
```
{
    "visualization_server": "gcr.io/ml-pipeline/visualization-server:2.0.3",
    "frontend": "gcr.io/ml-pipeline/frontend:2.0.3"
}
```
This makes the user experience error-prone because if you set in the `custom_images` config with the key `frontend_image` rather than `frontend`, it will not work.
This is the case for channels `latest/edge` and `2.0/stable`.

When setting the config as shown in the `To Reproduce` section below, the charm goes to error, that's not the expected behavior, it should rather log a warning and ignore the config. This happens due to issue #485  in `kfp-profile-controller` charm.

### To Reproduce

```
juju deploy kfp-profile-controller --trust --channel=2.0/stable

juju config kfp-profile-controller custom_images

#output
  custom_images: 
    default: |
      visualization_server : ''
      frontend_image : ''

# set the image as the default tells you
juju config kfp-profile-controller custom_images="visualization_server: 'gcr.io/ml-pipeline/visualization-server:2.0.3'
frontend_image: 'gcr.io/ml-pipeline/visualization-server:2.0.3'
"

# check juju status: charm goes to error
kfp-profile-controller/0*     error
```

### Environment

1.25-strict/stable
juju 3.1/stable

### Relevant Log Output

```shell
-
```

### Additional Context

_No response_

NohaIhab:
closes #484 

## Summary
The default of `custom_images` config in `kfp-profile-controller` charm says `frontend_image` whereas the actual key you should set is just `frontend` as defined in the `.json`

It is currently misleading to users and will cause them to set the config incorrectly, which will then not take effect as expected.

This PR sets the config default to the correct `frontend` value.

### DIFF
--- a/charms/kfp-profile-controller/config.yaml
+++ b/charms/kfp-profile-controller/config.yaml
@@ -4,9 +4,10 @@
 options:
   custom_images:
     type: string
+    # defaults should match src/defaults-custom-images.json
     default: | 
       visualization_server : ''
-      frontend_image : ''
+      frontend : ''
     description: >
       YAML or JSON formatted input defining images to use in Katib
       For usage details, see https://github.com/canonical/kfp-operators.


orfeas-k (approved):

NohaIhab:
the CI for `kfp-profile-controller` integration tests is failing due to #487
this is now blocked

NohaIhab:
CI is now unblocked by #499

Issue #492, opened 2024-06-03
Add logging relation to kfp-viewer charm 


rgildein:
### Context

Add logging relation using `loki_push_api` interface and [LogForwarder](https://github.com/canonical/loki-k8s-operator/blob/5fcf61697aad10f47c8a3df5bb9d31ac830dcc1a/lib/charms/loki_k8s/v1/loki_push_api.py#L356) to kfp-viewer charm. Alternatively the [LogProxyConsumer](https://github.com/canonical/loki-k8s-operator/blob/5fcf61697aad10f47c8a3df5bb9d31ac830dcc1a/lib/charms/loki_k8s/v1/loki_push_api.py#L224C4-L224C20), could be used if the service is using log file instead of STDOUT, however this needs to be discussed because LogProxyConsumer is [marked as deprecated](https://discourse.charmhub.io/t/promtail-based-log-streaming-to-loki-logproxyconsumer-is-scheduled-for-deprecation/14289).

This task is part of the COS integration initiative for all Kubeflow charms.

### What needs to get done

1. Add logging relation  using `loki_push_api` interface
2. Use LogForwarder or LogProxyConsumer
3. Use chisme abstraction for testing

### Definition of Done

1. Charm could be related with grafana-agent-k8s via logging relation
2. This relation was tested by integration tests
3. This relation was tested manually with COS deployed

PR #509, opened 2024-06-10
Add logging relation kfp-viewer

rgildein:
Added logging relation and cos integration tests from chisme.

fixes: #492

How I tested it:

from terminal
```bash
$ tox -e integration -- --model kubeflow --keep-models
$ juju consume u1-k8s:cos.loki-logging  # I have pre-deployed cos model
$ juju integrate loki-logging grafana-agent-k8s
```

I visit `http://<ip>/cos-grafana/explore` to explore the logs and see:
Screenshot from 2024-06-14 17-02-46


orfeas-k (commented):
Everything looks good. However, both in the PR's screenshot and in my local deployment, I see that logs are duplicated (while looking at the container's logs, there's only one instance of them). Could this be an issue on our side?

rgildein:
> Everything looks good. However, both in the PR's screenshot and in my local deployment, I see that logs are duplicated (while looking at the container's logs, there's only one instance of them). Could this be an issue on our side?

I tried it locally again and the behavior was always the same. I was not able to fix or track the error.
Furthermore, I found out that telco team hit a similar issue, and they track the issue to be due to updating pebble layer on update status event. The enhancement is described at TELCO-1130 task, and here is an [example](https://github.com/canonical/sdcore-amf-k8s-operator/pull/162) of how they fix it. I checked that in [chisme](https://github.com/canonical/charmed-kubeflow-chisme/blob/02f36173e81372d5681869dc674bf9cb60db816e/src/charmed_kubeflow_chisme/components/pebble_component.py#L139-L142) we are doing the same, so I'm not sure why we are seeing logs twice.

rgildein:
After some debugging, I found out that it was due passing integer for max_num_viewers, e.g. `max_num_viewers=self.model.config["max-num-viewers"],`.
This cause that during update-status hook, the services were different (see example below) and new layer was applied.

I add error log to this part of [chisme](https://github.com/canonical/charmed-kubeflow-chisme/blob/02f36173e81372d5681869dc674bf9cb60db816e/src/charmed_kubeflow_chisme/components/pebble_component.py#L139-L142) through juju ssh. And after update-status event, I was able to see that new layer was applied always since new layer has always `'MAX_NUM_VIEWERS': 25` defined as integer and current layer as string 'MAX_NUM_VIEWERS': '25'.
```bash
jhack fire kfp-viewer/0 update-status
unit-kfp-viewer-0: 16:41:20 ERROR unit.kfp-viewer/0.juju-log new_layer {'controller': Service({'summary': 'Entry point for kfp-viewer image', 'startup': 'enabled', 'override': 'replace', 'command': '/bin/controller -logtostderr=true -max_num_viewers=25 --namespace=', 'environment': {'MAX_NUM_VIEWERS': 25, 'NAMESPACE': ''}})}
unit-kfp-viewer-0: 16:41:20 ERROR unit.kfp-viewer/0.juju-log current_layer {'controller': Service({'summary': 'Entry point for kfp-viewer image', 'startup': 'enabled', 'override': 'replace', 'command': '/bin/controller -logtostderr=true -max_num_viewers=25 --namespace=', 'environment': {'MAX_NUM_VIEWERS': '25', 'NAMESPACE': ''}})}
```

Switching to `max_num_viewers=str(self.model.config["max-num-viewers"])` solved the issue and layer was applied always (during any event).  In provided screenshot is possible to see previous behavior and then the new one.

orfeas-k (approved):
Very good job on tackling down the duplication of logs! approving

PR #521, opened 2024-06-21
ci: use juju 3.4/stable instead of 3.5/stable

DnPlas:
Part of canonical/bundle-kubeflow#944


DnPlas:
Looks like this PR is affected by https://github.com/canonical/envoy-operator/issues/114.

PR #529, opened 2024-07-01
Change namespace in kfp-schedwf

### LINKED ISSUE #517: KFP Recurrings Runs not working on kubeflow 1.8
### Bug Description

I am experiencing an issue after a fresh installation of Kubeflow 1.8/stable (following the official guide). I can launch manual runs, and they execute successfully. However, recurring runs, whether they are Periodic or Cron, do not launch.

In contrast, if I install Kubeflow 1.7 with MicroK8s 1.24, recurring runs do work. Is anyone else experiencing the same issue?"

### To Reproduce

Fresh install (https://charmed-kubeflow.io/docs/get-started-with-charmed-kubeflow)

### Environment

Ubuntu 23.10
Microk8s 1.26
Charmed-kubeflow 1.8/stable

### Relevant Log Output

```shell
All units working as expected
```

### Additional Context

_No response_

mvlassis:
Closes #517 

This PR updates the `pebble_components.py` file in the kfp-schedwf charm to use a different namespace, similar to the one in [upstream](https://github.com/kubeflow/pipelines/blob/2.0.5/backend/Dockerfile.scheduledworkflow#L49).


charms/kfp-schedwf/src/components/pebble_component.py
DnPlas, line 1: Let's also remove [this](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-schedwf/src/charm.py#L78) line from the charm as it wont' be used.

This would also require us to remove the `namespace` arg from [here](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-schedwf/src/components/pebble_component.py#L17) and L28.

DnPlas (commented):
Thanks @mvlassis, tiny comment on the pebble component, also, how do you think we can improve the test coverage to prevent this in the future?

DnPlas (commented):
Thanks @mvlassis, this LGTM, but let's wait for #521 to be merged before approving this PR. It would be great to increase the test coverage to avoid this issue in the future.

Also, please apply this change and submit another PR for `track/2.0` as that version needs this fix more urgently. Let's approve that other PR first so we can release the fix asap.

mvlassis:
@DnPlas We can add an integration test for the `kfp-schedwf` charm, that creates a sample recurring run with a short trigger (a few seconds or minutes), and then check that the ScheduledWorkflow exists, and that the result of the run was successful.

DnPlas:
> @DnPlas To possibly catch these types of issues sooner, we can add an integration test for the `kfp-schedwf` charm. The test could create a sample recurring run with a short trigger (a few seconds or minutes), and then check that the ScheduledWorkflow exists, and that the result of the run was successful.

We have [this](https://github.com/canonical/kfp-operators/blob/main/tests/integration/test_kfp_functional_v2.py#L158) which makes sure the recurring run is there, but this test case is not monitoring the actual runs that are scheduled. Maybe we can extend it to check for the runs.

mvlassis:
> > @DnPlas To possibly catch these types of issues sooner, we can add an integration test for the `kfp-schedwf` charm. The test could create a sample recurring run with a short trigger (a few seconds or minutes), and then check that the ScheduledWorkflow exists, and that the result of the run was successful.
> 
> We have [this](https://github.com/canonical/kfp-operators/blob/main/tests/integration/test_kfp_functional_v2.py#L158) which makes sure the recurring run is there, but this test case is not monitoring the actual runs that are scheduled. Maybe we can extend it to check for the runs. wdyt?

We can extend the test to assert the result status of one (or more) runs, after the schedule is checked, and before the recurring runs are disabled, [here](https://github.com/canonical/kfp-operators/blob/1adef28ba22e41ff6d412c0f83a61c0c84e473e5/tests/integration/test_kfp_functional_v2.py#L192).

mvlassis:
I have now expanded the integration test `test_create_and_monitor_recurring_run` to check whether a run has been created and then assert that its status is `SUCCEEDED`

DnPlas (approved):
Amazing job, @mvlassis !

DnPlas (approved):

mvlassis:
Made some changes in `test_kfp_functional_v1.py` to incorporate differences in API versions.

mvlassis:
This PR is closed in favor of a (new one)[https://github.com/canonical/kfp-operators/pull/535] because one of the commits was not signed before it was commited.

PR #535, opened 2024-07-10
fix: watch for recurring runs in the correct namespace and expand integration tests for recurring runs

### LINKED ISSUE #517: KFP Recurrings Runs not working on kubeflow 1.8

mvlassis:
Closes #517 

This PR updates the kfp-schedwf charm to use an empty namespace, similar to the one in [upstream](https://github.com/kubeflow/pipelines/blob/2.0.5/backend/Dockerfile.scheduledworkflow#L49).

Also, the `create_and_monitor_recurring_run` test is expanded to actually test that the run has been created and finished successfully.

### DIFF
--- a/charms/kfp-schedwf/src/components/pebble_component.py
+++ b/charms/kfp-schedwf/src/components/pebble_component.py
@@ -14,7 +14,6 @@ class KfpSchedwfPebbleService(PebbleServiceComponent):
         *args,
         timezone: str,
         log_level: str,
-        namespace: str,
         **kwargs,
     ):
         """Pebble service container component in order to configure Pebble layer"""
@@ -22,8 +21,8 @@ class KfpSchedwfPebbleService(PebbleServiceComponent):
         self.environment = {
             "CRON_SCHEDULE_TIMEZONE": timezone,
             "LOG_LEVEL": log_level,
+            "NAMESPACE:": "",
         }
-        self.namespace = namespace
         self.log_level = log_level
 
     def get_layer(self) -> Layer:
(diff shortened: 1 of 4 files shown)


charms/kfp-schedwf/src/components/pebble_component.py
```diff
@@ -47,7 +46,7 @@ def get_layer(self) -> Layer:
                         "summary": "scheduled workflow controller service",
                         "startup": "enabled",
                         "command": f"/bin/controller --logtostderr=true"
```
NohaIhab, line 48: nit:
```suggestion
                        "command": "/bin/controller --logtostderr=true"
```

orfeas-k:
I would also add a more descriptive PR title (and thus commit title). "Fix namespace" doesn't provide a lot of insight on what this fix is when looking at the commit history

NohaIhab (approved):
LGTM, let's merge once the CI passes
thx @mvlassis ! can you also send a PR to track/2.0 to add the tests?

PR #541, opened 2024-07-19
Use charmcraft edge

misohu:


DnPlas:
@misohu should we close this issue now as we'll keep using latest/candidate?

DnPlas:
After #552 I don't think these changes are required, I suggest we close this issue.

Issue #549, opened 2024-07-29
CI is stuck running integration tests for almost two hours


DnPlas:
### Bug Description

Integration tests in the CI are failing to be completed; initially the `bundle-integration-v2` job appears to be stuck and the only logs we can see are:

```
bundle-integration-v2: commands[0]> pytest -vv --tb=native -s --model kubeflow --bundle=./tests/integration/bundles/kfp_latest_edge.yaml.j2 /home/runner/work/kfp-operators/kfp-operators/tests/integration/test_kfp_functional_v2.py
============================= test session starts ==============================
platform linux -- Python 3.8.10, pytest-8.3.2, pluggy-1.5.0 -- /home/runner/work/kfp-operators/kfp-operators/.tox/bundle-integration-v2/bin/python
cachedir: .tox/bundle-integration-v2/.pytest_cache
rootdir: /home/runner/work/kfp-operators/kfp-operators
configfile: pyproject.toml
plugins: operator-0.35.0, asyncio-0.21.2, anyio-4.4.0
asyncio: mode=strict
collecting ... collected 6 items
```

After almost two hours, the job fails and we get to see more logs. See from [L22](https://github.com/canonical/kfp-operators/actions/runs/10110429488/job/27960326357?pr=548#step:5:23).

Everything looks normal, the test cases are passing up until the ` tests/integration/test_kfp_functional_v2.py::test_viz_server_healthcheck` test case, which fails, BUT the logs from the failure do not indicate an error from `kfp-operators` or any other charm, what we see is:

```
  File "/home/runner/work/kfp-operators/kfp-operators/.tox/bundle-integration-v2/lib/python3.8/site-packages/juju/client/connection.py", line 649, in rpc
    await self._ws.send(outgoing)
  File "/home/runner/work/kfp-operators/kfp-operators/.tox/bundle-integration-v2/lib/python3.8/site-packages/websockets/legacy/protocol.py", line 635, in send
    await self.ensure_open()
  File "/home/runner/work/kfp-operators/kfp-operators/.tox/bundle-integration-v2/lib/python3.8/site-packages/websockets/legacy/protocol.py", line 948, in ensure_open
    raise self.connection_closed_exc()
websockets.exceptions.ConnectionClosedError: sent 1011 (internal error) keepalive ping timeout; no close frame received
websockets.exceptions.ConnectionClosedError: sent 1011 (internal error) keepalive ping timeout; no close frame received
------------------------------ Captured log call -------------------------------
[...]

PR #553, opened 2024-07-30
fix: increase juju MAX_FRAME_SIZE to avoid reconnection issues

DnPlas:
Increase Juju websocket connection MAX_FRAME_SIZE to 1024MiB to avoid RPC: Connection closed, reconnecting errors during integration test executions.

Part of #549


DnPlas:
Closing as this workaround did not work, I am still seeing  `ERROR    juju.client.connection:connection.py:662 RPC: Automatic reconnect failed` errors.

PR #560, opened 2024-08-15
Add alert rules to kfp-api based on the KF093 spec

rgildein:
These alert rules provide an overview of all service states.

Using the KubeflowServiceDown or KubeflowServiceIsNotStable filter, the user
can easily see the status of all Kubeflow services.

part-of: [#1026](https://github.com/canonical/bundle-kubeflow/issues/1026)


misohu (approved):
LGTM

PR #563, opened 2024-08-29
Update tools/get-images.sh to handle both Go and Python versions of yq

### LINKED ISSUE #562: tools/get-images.sh when not using snap version of `yq`
### Bug Description

Running the `tools/get-images.sh` with `bash tools/get-images.sh` produces a syntax error when running a version `yq` that hasn't been installed via snap.

The error occurs because there are 2 different maintained versions of `yq`, [one that uses Go](https://github.com/mikefarah/yq) and [one that uses Python](https://github.com/kislyuk/yq). While the Go version supports glob characters, the Python doesn't and produces this error message.

I noticed this because I tried running the script on an Arch Linux installation with `yq` installed via the package manager. However, this also occurs using Ubuntu, if you install `yq` with `pip` (because it installs the Python package).

### To Reproduce

- Clone the repository with `git clone` and `cd` to it
- Install the Python package of `yq` with `pip install yq`.
- Run 'bash tools/get-images.sh`

### Environment

- Ubuntu 22.04 when installing `yq` via `pip`
- Arch Linux

### Relevant Log Output

```shell
jq: error: syntax error, unexpected '*', expecting FORMAT or QQSTRING_START or '[' (Unix shell quoting issues?) at <top-level>, line 1:
.options.*-image |  select(.) | .default         
jq: 1 compile error
```

### Additional Context

To check the version of `yq`, you can run `yq --version` and inspect the output.
When using the Go version, the output should be similar to:
```
yq (https://github.com/mikefarah/yq/) version v4.44.2
```

When using the Python version, the output should be similar to:
```
yq 3.4.3
```

mvlassis:
Closes #562

This PR updates `tools/get-images.sh` to be version agnostic when it comes to the `yq` utility. There is additional context in the [issue's description](https://github.com/canonical/kfp-operators/issues/562#issue-2492407938).

From experimentation, the Python version of `yq` also adds quotes to the output, so `tr` is called to remove any quote characters from all the entries.

To test:
- Run `bash tools/get-images.sh` with the snap version of `yq` (installed via `sudo snap install yq`)
- Run `bash tools/get-images.sh` with the Python verison of `yq` (installed via `pip install yq`)
- The output should be the same, and the same as when running `bash tools/get-images.sh` in the `main` branch.

### DIFF
--- a/tools/get-images.sh
+++ b/tools/get-images.sh
@@ -3,6 +3,9 @@
 # This script returns list of container images that are managed by this charm and/or its workload
 #
 # dynamic list
+
+set -xe
+
 IMAGE_LIST=()
 IMAGE_LIST+=($(find -type f -name metadata.yaml -exec yq '.resources | to_entries | .[] | .value | ."upstream-source"' {} \; | tr -d '"'))
 IMAGE_LIST+=($(find -type f -name config.yaml -exec yq '.options | with_entries(select(.key | test("-image$"))) | .[].default' {} \; | tr -d '"'))
--- a/tools/get-images.sh
+++ b/tools/get-images.sh
@@ -4,6 +4,6 @@
 #
 # dynamic list
 IMAGE_LIST=()
-IMAGE_LIST+=($(find -type f -name metadata.yaml -exec yq '.resources | to_entries | .[] | .value | ."upstream-source"' {} \;))
-IMAGE_LIST+=($(find -type f -name config.yaml -exec yq '.options.*-image |  select(.) | .default' {} \;))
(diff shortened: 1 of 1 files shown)


rgildein (dismissed):
LGTM,
tested versions:
- yq (`v4.44.2 (2566)`) installed with snap
- yq (`Version: 3.4.3`) installed with pip

kimwnasptd:
(discussed during the daily)

The CI is currently blocked because of https://github.com/canonical/charmcraft/issues/1845

kimwnasptd (approved):

PR #565, opened 2024-09-02
Pin charmcraft channel to `3.1/stable`

mvlassis:
This PR is part of closing [this issue](https://github.com/canonical/bundle-kubeflow/issues/1049).

It pins the channel used in our CI from `latest/candidate` to `latest/stable`.

Ideally we can unpin the version when [this issue](https://github.com/canonical/charmcraft/issues/1845) is fixed.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -98,6 +98,7 @@ jobs:
           provider: microk8s
           channel: 1.29-strict/stable
           juju-channel: 3.4/stable
+          # Pinned to 3.1/stable due to https://github.com/canonical/charmcraft/issues/1845
           charmcraft-channel: 3.1/stable
 
       - name: Integration tests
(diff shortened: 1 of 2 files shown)


.github/workflows/integrate.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
           provider: microk8s
           channel: 1.29-strict/stable
           juju-channel: 3.4/stable
-          charmcraft-channel: latest/candidate
+          charmcraft-channel: 3.1/stable
```
DnPlas, line 101: Please add a comment that says why we are pinning this version and when we should unpin, for example:

```suggestion
          # Pinned to 3.1/stable due to https://github.com/canonical/charmcraft/issues/1845
          charmcraft-channel: 3.1/stable
```

DnPlas (commented):
Thanks @mvlassis, tiny nit.

.github/workflows/integrate.yaml
mvlassis: Good practice, thanks!

rgildein (approved):
LGTM

PR #583, opened 2024-11-13
chore: Upgrade manifests to 2.3.0

### LINKED ISSUE #580: Upgrade the manifests for 2.3.0
### Context

This is a follow-up from https://github.com/canonical/kfp-operators/issues/579

To have a full 2.3.0 upgrade the first part will be to update all the manifests that get deployed by the charms, to update from 2.2.0 to 2.3.0.

Although with a quick look on the diff between the upstream manifests I see differences only in the OCI images that are used https://github.com/canonical/kfp-operators/issues/579#issuecomment-2473040282

### What needs to get done

1. Look into upstream and downstream manifests and decide what updates we need to make
2. Update the manifests that get applied by the charms, if needed

### Definition of Done

1. Manifests applied by the charms are in sync with 2.3.0

orfeas-k:
Upgrade manifests to 2.3.0. For producing the manifests, instructions in the [CONTRIBUTING.md](https://github.com/canonical/kfp-operators/blob/main/CONTRIBUTING.md) file were followed.

This includes:
* Upgrading images
* Updating kfp-api auth-manifests
* Introducing a new envvar in kfp-ui

Fixes #580

#### Tests
CI is fails mostly intermittently due to #602 and #601. Locally, they are successful (I hit #601 only once which looks unrelated to charms changes). I 've also refreshed a CKF deployment with the PR's charms and ran some pipelines, the kfp UAT and also enabled a recurring one.

### DIFF
--- a/charms/kfp-api/config.yaml
+++ b/charms/kfp-api/config.yaml
@@ -51,13 +51,13 @@ options:
     description: Default name of object storage bucket.
   launcher-image:
     type: string
-    # Source: https://github.com/kubeflow/pipelines/blob/2.2.0/backend/src/v2/compiler/argocompiler/container.go#L30
-    default: "gcr.io/ml-pipeline/kfp-launcher@sha256:8fe5e6e4718f20b021736022ad3741ddf2abd82aa58c86ae13e89736fdc3f08f"
+    # Source: https://github.com/kubeflow/pipelines/blob/2.3.0/backend/src/v2/compiler/argocompiler/container.go#L33
+    default: "gcr.io/ml-pipeline/kfp-launcher@sha256:bef55a344574a25c557256d7c66cb19edacfd2008d694e5b6bb5b612d59feae0"
     description: Launcher image used during a pipeline's steps.
   driver-image:
     type: string
-    # Source: https://github.com/kubeflow/pipelines/blob/2.2.0/backend/src/v2/compiler/argocompiler/container.go#L32
-    default: "gcr.io/ml-pipeline/kfp-driver@sha256:3c0665cd36aa87e4359a4c8b6271dcba5bdd817815cd0496ed12eb5dde5fd2ec"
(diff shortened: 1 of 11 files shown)


DnPlas (commented):
All changes LGTM, for testing this I deployed all kfp charms and checked the status, as well as their logs.
Since `kfp-ui` is the charm that received the most changes, I focused on it and inspected the pebble logs. While most of it looks normal, I noticed two things:

1. There is a `2024-11-20T04:12:03.662Z [ml-pipeline-ui] Failed reading json data from '/etc/config/viewer-pod-template.json':
2024-11-20T04:12:03.664Z [ml-pipeline-ui] SyntaxError: Unexpected end of JSON input` message.
2. There are a lot of warnings like `2024-11-20T04:12:03.686Z [ml-pipeline-ui] (node:15) Warning: Accessing non-existent property 'which' of module exports inside circular dependency`

I don't think any of this is affecting the charm, but it's worth double checking.

orfeas-k:
Hey @DnPlas, thank you for taking a thorough look. For the first one, I have already filed this issue https://github.com/canonical/kfp-operators/issues/603. This has been observed in `latest/edge` kfp-ui (without the upgrades) and I think it should be present in 1.9 too. Regarding 2, this is also present in `latest/edge` without the upgrades. Let me file an issue for this as well and we 'd need to investigate separately

NohaIhab (changes requested):
Thanks @orfeas-k , `bundle-integration-v2` tests all passed locally
I can confirm the logs reported by @DnPlas do not affect the functionality

one thing missed: to upgrade the driver and launcher images in `config.yaml` of `kfp-api` charm

orfeas-k:
@NohaIhab Very good catch Noha. Updated the PR. Also, I 've opened #610 against the `dev` branch updating the contributing.md file.

DnPlas (approved):
LGTM, thanks @orfeas-k

NohaIhab (approved):

Issue #639, opened 2024-12-17
Use `.charm` artefacts in individual integraton tests


DnPlas:
### Context

Based on the changes proposed in https://github.com/canonical/kfp-operators/issues/635 and https://github.com/canonical/bundle-kubeflow/issues/1070, and assuming the changes introduced by https://github.com/canonical/charmed-kubeflow-workflows/pull/88, the individual integration tests must:

* Download the `charm` artefacts generated by the [`get_charms_build_with_cache.yaml`](https://github.com/canonical/charmed-kubeflow-workflows/pull/88/files#diff-53a551a14277df619fdf1fb5a1f92d7a14b409f6a1e5bd4c8f57b6bdc6c9c69d) reusable workflow and use them for deploying them and run tests on top
* Not build the charms

### What needs to get done

1. Allow the `integration` job to download the `charm` artefacts
2. Remove the build step from every individual integration test
3. Use the `charm` artefacts for deploying charms under test

### Definition of Done

Individual integration tests use `charm` artefacts to deploy and test them.

PR #645, opened 2025-01-06
ci: use get_charms_build_with_cache.yaml workflow and use charm artefacts for individual tests

### LINKED ISSUE #641: Use `get_charms_build_with_cache.yaml` on pull request workflow
### Context

https://github.com/canonical/charmed-kubeflow-workflows/pull/88 introduces a reusable workflow that allows the CI to build charms in their dedicated machines and with charmcraft cache enabled. These enablements improve the build time and ensure that the same charm artefact is tested individually, as a bundle, and published.

### What needs to get done

1. Add the canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml workflow on pull request workflow

```yaml
name: On Pull Request

# On pull_request, we:
# * always publish to charmhub at latest/edge/branchname
# * always run tests
# * always run builds with cache
on:
  pull_request:

jobs:
  get-paths-and-build:
    name: Get charm paths and build with cache
    uses: canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml@main

  tests:
    name: Run Tests
    needs: [get-paths-and-build]
    uses: ./.github/workflows/integrate.yaml
    secrets: inherit

  # publish runs in parallel with tests, as we always publish in this situation
  publish-charm:
    name: Publish Charm
    needs: [get-paths-and-build]
    uses: ./.github/workflows/publish.yaml
    secrets: inherit
```

### Definition of Done

The pull request workflow uses canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml before proceeding on the publish and test jobs.

DnPlas:
This PR merges `KF-6684-refactor-ci-with-build` branch into `main`. These are the changes:

* ci: use get_charms_build_with_cache.yaml workflow

This workflow enables the usage of canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml for this CI, meaning that the integration tests and publish jobs are now able to download charm artefacts and use them consistently throught the various operations of this CI.

Fixes #641

* ci, tests: pass charm artefacts to deploy and test charms

This commit enables the "--charm-path" option to pass .charm artefacts
to be deployed and tested at an individual level.
This change enables the option to pass pre-built charms to the tests
to avoid building them on the same test. It also ensures compatibility
with the build_charm.py reusable workflow (from canonical/data-platform-workflows).

Fixes: https://github.com/canonical/kfp-operators/issues/639


.github/workflows/on_pull_request.yaml
```diff
@@ -3,19 +3,24 @@ name: On Pull Request
...
 # * always run tests
-
+# * always run builds with cache
 on:
   pull_request:
 
 jobs:
+  get-paths-and-build:
+    name: Get charm paths and build with cache
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml@main
```
NohaIhab, line 13: I'm concerned about how we will version this workflow, and whether it will be supported by data platform team as they are the ones that maintain the workflow reused in https://github.com/canonical/charmed-kubeflow-workflows/blob/main/.github/workflows/get_charms_build_with_cache.yaml, pinging @carlcsaposs-canonical

DnPlas: > I'm concerned about how we will version this workflow

Hey @NohaIhab, this workflow only wraps an action for getting charm paths and the build_charm.yaml reusable workflow, so I don't think a versioning system other than pointing to `main` to be that necessary. Were you thinking of something else? Is there a concern with it knowing this information?

Thanks @carlcsaposs-canonical for your reply and insight, a couple comments:

> I would strongly recommend against wrapping the data-platform-workflows inside of other reusable workflows, unless there's a compelling use case for it across multiple repositories

The charmed-kubeflow reusable workflow contains an action that gets charm paths, and then passes them to the build_charm.yaml workflow, this way we avoid some boilerplate in multiple repositories.

> pinning a version (of the workflow) for several repos in one place can cause a lot of headaches
also pinning a version (of the workflow) for multiple commits/branches of one repo in one place can cause a lot of headaches 

I think this can be resolved by exposing some workflow options, for starters both the tools versions and the version of the workflow itself.

> data-platform-workflows provides both reusable workflows and python dependencies (namely [pytest plugins](https://github.com/canonical/data-platform-workflows/tree/main/python/pytest_plugins)). you're not using pytest plugins right now, but if you plan to in the future (e.g. for allure dashboards to track ci stability/speed), the pytest plugins and workflows should be versioned & updated together. (data-platform-workflows only supports using one version for all components in a repo, so if the workflow is v24.0.2, the python plugin needs to be v24.0.2 or there may be compatibility issues)

That's a good point. I think it can still be solved by exposing the workflow version in the wrapper, no? That way if a repository uses the pytest plugin, we'll always have to pin the version and update it in the same PR according to the update suggestions.

> some benefits of standardization that would be lost:
using the same approach for versioning/updates allows knowledge sharing/transfer, especially in troubleshooting common issues

Do you think this will be lost even if we expose the version of the wf?

> making changes to data-platform-worfklows becomes more difficult—for example, for some of the breaking changes to data-platform-workflows (e.g. the charmcraft 3 migration), I make an effort to open PRs across all our repos to keep everyone up-to-date—if the reusable workflows are called directly, I'd be happy to extend that to the kubeflow repos as well, but if there's an extra layer of abstraction it would make that much more difficult so I wouldn't be able to do that

For this one I'm not really able to see what the blocker would be. I have in mind #643, which changes the charms metadata directly, and does very minimal changes to the workflows. In fact, this is not even touching the build_charm.yaml workflow, so I wonder what sort of changes could there be that will make this task any different.

---

In the end, the wrapper is just avoiding us the get charm paths boilerplate, if we think the benefits of using it are rather small compared to the way the data platform team is doing it, we can change it, but I cannot seem to find something that's very troublesome. Some of the benefits of keeping the wrapper:

* Keep the version centralised, so one update for the 30+ repositories that we have
* Avoid the boilerplate of getting charm paths
* Any fix we'll have to make will happen in one repo, rather than in 30+, unless we need to pin a specific version (e.g. charmcraft, lxd), BUT we've rarely seen this, usually if we have to pin a version in one repo, it has to be replicated in ALL repos.


@NohaIhab @orfeas-k thoughts?

NohaIhab: resolved in #657

Issue #650, opened 2025-01-16
Use `build_charm` workflow from data platform in `On Push` workflow


NohaIhab:
### Context

Similarly to what's implemented in #645 for `On Pull Request` workflow, the same should be applied for `On Push`. That is use `build_charm.yaml` workflow.

### What needs to get done

refactor the on push workflow to use `build_charm.yaml` 

### Definition of Done

on push uses `build_charm.yaml`

NohaIhab:
Closed by #657

PR #651, opened 2025-01-16
ci: remove wrapper around get charm paths and build with cache workflows

DnPlas:
This commit removes the extra abstraction of the get-charm-paths action and build_charm.yaml reusable workflow to avoid potential incompatibilities and sync issues with the data platforms workflows.

Notes to reviewer:
* the bundle tests will be addressed in #654 
* the on push workflow is now broken due to the changes, and it will be addressed in #650 
* the release and publish workflow will be addressed in #656

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -134,7 +134,7 @@ jobs:
         timeout-minutes: 5
         uses: actions/download-artifact@v4
         with:
-          pattern: ${{ inputs.artifact-prefix }}-${{ matrix.charm }}-*
+          pattern: ${{ inputs.artifact-prefix }}-*
           merge-multiple: true
 
       - name: Integration tests
@@ -145,7 +145,7 @@ jobs:
           # Pass the path where the charm artefact is downloaded to the tox command
           # FIXME: Right now the complete path is half hardcoded to <charm name>_ubuntu-20.04-amd64.charm
           # We need to find a better way to dynamically get this value
-          sg snap_microk8s -c "tox -e ${{ matrix.charm }}-integration -- --model kubeflow --charm-path=${{ github.workspace }}/charms/${{ matrix.charm }}/${{ matrix.charm }}_ubuntu-20.04-amd64.charm"
+          sg snap_microk8s -c "tox -e ${{ matrix.charm }}-integration -- --model kubeflow --charm-path=${{ github.workspace }}/charms/${{ matrix.charm }}/${{ matrix.charm }}_ubuntu@20.04-amd64.charm"
 
       - name: Collect charm debug artifacts
         uses: canonical/kubeflow-ci/actions/dump-charm-debug-artifacts@main
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -6,6 +6,14 @@ on:
     secrets:
       CHARMCRAFT_CREDENTIALS:
         required: true
+    inputs:
+      artifact-prefix:
+        description: |
+          Prefix for charm package GitHub artifact(s)
+
+          Use canonical/data-platform-workflows build_charm.yaml to build the charm(s)
+        required: true
+        type: string
 
 jobs:
   lib-check:
@@ -126,7 +134,7 @@ jobs:
         timeout-minutes: 5
         uses: actions/download-artifact@v4
         with:
-          pattern: packed-charm-cache-true-.-charms-${{ matrix.charm }}-*
+          pattern: ${{ inputs.artifact-prefix }}-${{ matrix.charm }}-*
           merge-multiple: true
 
       - name: Integration tests
--- a/.github/workflows/on_pull_request.yaml
+++ b/.github/workflows/on_pull_request.yaml
@@ -38,6 +38,8 @@ jobs:
     needs: build
     uses: ./.github/workflows/integrate.yaml
     secrets: inherit
+    with:
+      artifact-prefix: ${{ needs.build.outputs.artifact-prefix }}
 
   # publish runs in parallel with tests, as we always publish in this situation
   publish-charm:
--- a/.github/workflows/on_pull_request.yaml
+++ b/.github/workflows/on_pull_request.yaml
@@ -8,9 +8,31 @@ on:
   pull_request:
 
 jobs:
-  get-paths-and-build:
-    name: Get charm paths and build with cache
-    uses: canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml@main
+  get-charm-paths:
+    name: Generate the Charm Matrix content
+    runs-on: ubuntu-latest
+    outputs:
+      charm_paths: ${{ steps.get-charm-paths.outputs.charm-paths }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Get paths for all charms in this repo
+        id: get-charm-paths
+        uses: canonical/kubeflow-ci/actions/get-charm-paths@main
+
+  build:
+    name: Build charms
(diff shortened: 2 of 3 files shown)


.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Get paths for all charms in this repo
+        id: get-charm-paths
+        uses: canonical/kubeflow-ci/actions/get-charm-paths@main
+
+  build:
+    name: Build charms
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v24.0.5
```
NohaIhab, line 26: let's use v29 now to test with the latest changes

.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
...
+    needs: get-charm-paths
+    strategy:
+      fail-fast: false
+      matrix:
+        charm: ${{ fromJson(needs.get-charm-paths.outputs.charm_paths) }}
+    with:
+      cache: true 
+      charmcraft-snap-channel: 3.x/stable
```
NohaIhab, line 34: `latest/candidate` is available for 3.3, I suggest to use it here and add a comment to remove it after charmcraft 3.3 stable release

NohaIhab (commented):
thanks @DnPlas, a couple comments on changes for the workflow and charmcraft versions. I suggest @carlcsaposs-canonical to also give it a review to ensure we are reusing their workflow as expected.

.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
...
+      path-to-charm-directory: ./${{ matrix.charm }}
```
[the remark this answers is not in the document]
DnPlas: Updated

.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
...
+      - name: Get paths for all charms in this repo
+        id: get-charm-paths
+        uses: canonical/kubeflow-ci/actions/get-charm-paths@main
+
+  build:
+    name: Build charms
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v24.0.5
+    needs: get-charm-paths
+    strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
DnPlas: Removed

.github/workflows/on_pull_request.yaml
DnPlas: Updated

.github/workflows/on_pull_request.yaml
DnPlas: Updated

.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
...
+        charm: ${{ fromJson(needs.get-charm-paths.outputs.charm_paths) }}
+    with:
+      cache: true 
+      charmcraft-snap-channel: 3.x/stable
+      path-to-charm-directory: ./${{ matrix.charm }}
 
   tests:
     name: Run Tests
-    needs: [get-paths-and-build]
+    needs: build
```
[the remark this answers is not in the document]
NohaIhab: yes it's valid, if it's just one job it doesn't need to be a list. See [documentation](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

.github/workflows/integrate.yaml
```diff
@@ -6,6 +6,14 @@ on:
...
       CHARMCRAFT_CREDENTIALS:
         required: true
+    inputs:
+      artifact-prefix:
+        description: |
+          Prefix for charm package GitHub artifact(s)
+
+          Use canonical/data-platform-workflows build_charm.yaml to build the charm(s)
+        required: true
+        type: string
```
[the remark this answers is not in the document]
NohaIhab: After discussing we will be re-using the same cached builds for the release workflow in #656 , so let's keep it untouched for now, same applies for the publish.yaml. Thanks @carlcsaposs-canonical for the helpful reviews!

.github/workflows/integrate.yaml
```diff
@@ -6,6 +6,14 @@ on:
     secrets:
       CHARMCRAFT_CREDENTIALS:
         required: true
+    inputs:
+      artifact-prefix:
+        description: |
+          Prefix for charm package GitHub artifact(s)
+
+          Use canonical/data-platform-workflows build_charm.yaml to build the charm(s)
```
mvlassis, line 14: We could point to the actual `build_charm.yaml` here: https://github.com/canonical/data-platform-workflows/blob/main/.github/workflows/build_charm.yaml

.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
...
-    uses: canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml@main
+  get-charm-paths:
+    name: Generate the Charm Matrix content
+    runs-on: ubuntu-latest
+    outputs:
+      charm-paths: ${{ steps.get-charm-paths.outputs.charm-paths }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
mvlassis, line 19: I'm not sure I understand why we use `fetch-depth: 0` in this case, if we don't use any other branch

.github/workflows/on_pull_request.yaml
```diff
@@ -8,19 +8,41 @@ on:
   pull_request:
 
 jobs:
-  get-paths-and-build:
-    name: Get charm paths and build with cache
-    uses: canonical/charmed-kubeflow-workflows/.github/workflows/get_charms_build_with_cache.yaml@main
+  get-charm-paths:
+    name: Generate the Charm Matrix content
+    runs-on: ubuntu-latest
```
mvlassis, line 13: Is using `ubuntu-latest` OK in this case? We usually pin to a specific version

mvlassis (commented):
Very small notes. 

suggestion: We could update the `actions/checkout` step to `actions/checkout@v4` since we're updating the workflows.

.github/workflows/on_pull_request.yaml
NohaIhab: If we pin to a certain version, we'd have to implement a strategy to update it. I don't see a reason here why we should pin. What do you think?

.github/workflows/integrate.yaml
NohaIhab: I think we can link the repo if needed, the link to the file would be tied to a certain branch, that can be changed in L31 of `on_pull_request.yaml` , so it would be difficult to keep this comment up to date if we link to the file

NohaIhab: Although IMO `canonical/data-platform-workflows build_charm.yaml` is clear enough

.github/workflows/on_pull_request.yaml
NohaIhab: It's in the [Example usage](https://github.com/canonical/kubeflow-ci/tree/main/actions/get-charm-paths#example-usage). I'm not sure if it this is expected to be the case, but if we need to use anything other than the latest commit it is needed.

mvlassis: It's still not clear for me why that is needed, we can specify the `ref` with the `ref` field:
```
steps:
  - name: Checkout specific commit
    uses: actions/checkout@v4
    with:
      ref: <commit-SHA>
```
But that is not related to this commit, I'll try to investigate why we need this

.github/workflows/integrate.yaml
mvlassis: Leaving it up to you :)

PR #677, opened 2025-02-14
fix: use charmcraft promote in promote.yaml workflow

DnPlas:
This commit rewrites the promote.yaml workflow to use charmcraft promote command instead of the release-charm action.We can no longer use the release-charm action because it relies on having GH releases in the repo, which we have stopped producing since migrating our CI to data-platform-workflows

Part of canonical/bundle-kubeflow#1202


NohaIhab (approved):
thx @DnPlas !

PR #678, opened 2025-02-18
[KF-7014] Use open requirements on terraform juju provider version sp…

deusebio:
…ecification

### DIFF
--- a/charms/kfp-api/terraform/versions.tf
+++ b/charms/kfp-api/terraform/versions.tf
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     juju = {
       source  = "juju/juju"
-      version = "~> 0.14.0"
+      version = ">= 0.14.0"
     }
   }
 }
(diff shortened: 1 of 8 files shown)


DnPlas (approved):
Thanks @deusebio ! The CI is failing, but it is possible to merge this PR in red.

#### Why is the CI failing?

1. While most of the integration tests seem to have failed, the reason they did was actually a failure in the `Collect charm debug artifacts` step. The logs say:

```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

This issue was tracked in https://github.com/canonical/kubeflow-ci/issues/152, but the fixes for it targeted `main` rather than the branches. In the future we can consider backporting the fix to the supported branches.

2. The `Upload charm to charmhub` step is failing for `kfp-metadata-writer`, but this was caused by a connection issue (nothing related to the changes introduced by this PR).

````
charmcraft internal error: NetworkError("('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))")
```

That being said, and because this PR is only changing the juju TF provider version constraints, the failures are not at all related to the changes introduced by this PR.

PR #706, opened 2025-05-05
fix(workaround): set constraints to setuptools-scm to avoid install loops

orfeas-k:
Due to python/importlib_metadata#516, we need to set up constaints
to the `setuptools-scm` to avoid install loops for these charms dependencies,
as recommended by the charmcraft team (see https://github.com/canonical/charmcraft/issues/2259#issuecomment-2842766428).

Fixes #293


mvlassis (approved):
Approved!

PR #711, opened 2025-05-06
feat(tests): Move charms deployed during tests to separate file

orfeas-k:
Move charms deployed during tests to a separate file called charms_dependencies.py. This enables programmatic access to them and thus updating them centrally, probably with the use of `kfcicli`. This uses `CharmSpec` dataclass from chisme to define dependency charms as separate objects. Also, a helper function enables generating context for cases where a bundle.yaml is used during tests.

Ref canonical/bundle-kubeflow/issues/1256

#### Note to reviewer
This is using common code introduced in https://github.com/canonical/charmed-kubeflow-chisme/pull/138. This should probably be reviewed first or maybe at the same time to get a more holistic understanding. Regarding changes in this repo:
* All charms have the same logic, so if we 're good with one, then we 're good with all.
* The only difference is cases where we use a `bundle.yaml` file to deploy charms. Those cases are `kfp-ui` tests and bundle integration tests.


charms/kfp-api/tests/integration/charms_dependencies.py
NohaIhab, line 1: I see what you are doing here with the dict now, relating to my [comment](https://github.com/canonical/charmed-kubeflow-chisme/pull/138#discussion_r207713792). I thinkk we can define the charms here straight away, that's what I had in mind. For example:
```
KFP_VIZ = CharmSpec(charm="kfp-viz", channel="latest/edge", trust=True)
```
and in the test file:
```
from charms_dependencies import KFP_VIZ
```
but maybe that does not align with kfcicli (?) I might be missing some context. Let's keep the conversation in the chisme PR then come back to this.

tests/integration/bundles/bundle.yaml.j2
NohaIhab, line 1: praise: nice job refactoring this directory!

NohaIhab (commented):
Hi @orfeas-k, left a comment and I think we need to resolve the chisme PR first before continuing the review. This was very helpful to visualize the use of the dataclass and helper though!

orfeas-k:
Force pushed to rebase and resolve commits but changed files are the same.

charms/kfp-api/tests/integration/charms_dependencies.py
orfeas-k: Pushed a commit refactoring to having dependencies as separate CharmSpec objects and using the new helper function that takes a List as an argument.

NohaIhab (approved):
Thank you @orfeas-k, really nice contribution for standardizing test dependencies!

PR #726, opened 2025-06-27
chore: Re-enable usage of cache while bulding charms

mvlassis:
Ref: https://github.com/canonical/bundle-kubeflow/issues/1290

This PR re-enables the usage of cache in all `multi-charm` repos.


orfeas-k (commented):
Changes look good but I see that the CI is broken for some reason. Could we investigate a bit more and have an issue?

mvlassis:
It seems that was an intermittent issue (the `bundle` tests not working)

orfeas-k (approved):

PR #728, opened 2025-07-01
feat: refactor rocks in the `kfp-profile-controller` unit tests

misohu:
Part of: https://github.com/canonical/bundle-kubeflow/issues/1272 

Removes the specific images from unit tests. Now we load them from `default-custom-images.json`

### DIFF
--- a/charms/kfp-profile-controller/tests/unit/test_operator.py
+++ b/charms/kfp-profile-controller/tests/unit/test_operator.py
@@ -1,6 +1,7 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+import json
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


NohaIhab (approved):
LGTM

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -19,6 +21,13 @@
     KfpProfileControllerOperator,
 )
 
+# Load the custom images from the JSON
+CUSTOM_IMAGES_PATH = (
+    Path(__file__).resolve().parent.parent.parent / "src" / "default-custom-images.json"
```
orfeas-k, line 26: I think you don't need the `.resolve().parent.parent.parent`. Here's how we load the metadata file in some other tests https://github.com/canonical/kfp-operators/blob/2a72102ea9f42f99f60983afd8759a2de3cf215e/charms/kfp-api/tests/integration/test_charm.py#L22.

NohaIhab (approved):

Issue #740, opened 2025-07-17
Upgrade Manifests to 2.5


MattiaSarti:
### Context

An upgrade of KFP to 2.5 is required.

### What needs to get done

1. for each charm, compare the differences between the new version of its upstream manifest and the current version of the downstream one
2. if required, modify each charm's manifest to have it in sync with the new upstream version

### Definition of Done

1. all charms' manifests are in sync with the upstream 2.5 ones

PR #742, opened 2025-07-18
chore: upgrade rocks to 2.5.0

MattiaSarti:
resolves https://github.com/canonical/kfp-operators/issues/741

NOTE: this pull request also reverts the change previously made to the file `charms/kfp-profile-controller/src/components/pebble_components.py` by https://github.com/canonical/kfp-operators/pull/745, which was a commented-out line never tested to run the Pebble services of `kfp-profile-contoller` as user `_deamon_`, because the [employed workload rock](https://hub.docker.com/r/ubuntu/python):
- is not configured to run a `_deamon_` user
- is already run as non-root, as witnessed by inspecting the used tag:
  ```bash
  docker inspect docker.io/ubuntu/python:3.12-24.04_stable --format='{{.Config.User}}'
  584792
  ```

PR #745, opened 2025-07-23
chore: upgrade charms' manifests

MattiaSarti:
resolves https://github.com/canonical/kfp-operators/issues/740

NOTE: while the manifests update Envoy's image too, there are [no upstream changes for its image build](https://github.com/kubeflow/pipelines/tree/2.5.0/third_party/metadata_envoy), so no Envoy changes were included in this pull request.

NOTE: the _**old downstream**_ images were temporarily switched to the _**new upstream**_ ones and will eventually be replaced by the _**new downstream**_ ones as in https://github.com/canonical/kfp-operators/pull/742 after this pull request is tested and approved.

### DIFF
--- a/charms/kfp-persistence/src/charm.py
+++ b/charms/kfp-persistence/src/charm.py
@@ -56,25 +56,6 @@ class KfpPersistenceOperator(CharmBase):
             depends_on=[self.leadership_gate],
         )
 
-        self.kubernetes_resources = self.charm_reconciler.add(
-            component=KubernetesComponent(
-                charm=self,
-                name="kubernetes:auth",
-                resource_templates=K8S_RESOURCE_FILES,
-                krh_resource_types={ClusterRole, ClusterRoleBinding, ServiceAccount},
-                krh_labels=create_charm_default_labels(
-                    self.app.name, self.model.name, scope="auth"
-                ),
-                context_callable=lambda: {
-                    "app_name": self.app.name,
-                    "namespace": self.model.name,
-                    "sa_name": SA_NAME,
-                },
-                lightkube_client=lightkube.Client(),
-            ),
-            depends_on=[self.leadership_gate],
-        )
-
         # creating a serviceAccountToken injected via a mounted projected volume:
         self.sa_token = self.charm_reconciler.add(
             component=SaTokenComponent(


charms/kfp-api/src/templates/auth_manifests.yaml.j2
```diff
@@ -1,109 +1,6 @@
...
-  name: {{ app_name }}-sa
-  namespace: {{ namespace }}
----
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  labels:
-    app: {{ app_name }}
-  name: {{ app_name }}-sa
-  namespace: {{ namespace }}
```
NohaIhab, line 106: suggestion(blocking): I'm +1 to removing these but we should keep the PRs atomic, I suggest we create an issue describing why the roles are not needed and send a separate PR from this one removing them.

charms/kfp-metadata-writer/tests/unit/test_operator.py
```diff
@@ -46,7 +46,7 @@ def test_kubernetes_component_created(harness, mocked_lightkube_client):
 
     # Assert that expected amount of apply calls were made
     # This simulates the Kubernetes resources being created
-    assert mocked_lightkube_client.apply.call_count == 3
```
NohaIhab, line 49: If removing the auth manifests means this charm is no longer creating k8s resources, then we should completely remove the k8s component. Note that it should be done in the roles removal PR as per my other comment.

charms/kfp-schedwf/src/charm.py
```diff
@@ -58,12 +65,30 @@ def __init__(self, *args):
...
+                    "app_name": self.app.name,
+                    "namespace": self.model.name,
+                    "sa_name": SA_NAME,
+                },
                 lightkube_client=lightkube.Client(),
             ),
             depends_on=[self.leadership_gate],
         )
 
+        self.sa_token = self.charm_reconciler.add(
```
NohaIhab, line 78: suggestion(non-blocking): add a comment explaining what this component maps to in upstream

charms/kfp-api/src/templates/auth_manifests.yaml.j2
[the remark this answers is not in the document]
NohaIhab: On top of the CRDs, 2 webhookConfiguration objects were addded: 
1. [MutatingWebhookConfiguration](https://github.com/kubeflow/manifests/blob/v1.10.1/apps/pipeline/upstream/base/webhook/pipelineversion-mutating-webhook-config.yaml)
2. [ValidatingWebhookConfiguration](https://github.com/kubeflow/manifests/blob/v1.10.1/apps/pipeline/upstream/base/webhook/pipelineversion-validating-webhook-config.yaml)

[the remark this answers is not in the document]
NohaIhab: I see that the above resources are not included in the default installation i.e. when I generate the manifests following [upstream instructions](https://github.com/kubeflow/manifests/blob/v1.10.1/README.md#kubeflow-pipelines) they are not there, not sure if this is intentional upstream. Think it's worth checking.

charms/kfp-ui/tests/unit/test_operator.py
```diff
@@ -59,7 +59,7 @@ def test_kubernetes_created_method(
     harness.charm.on.install.emit()
 
     # Assert
-    assert mocked_lightkube_client.apply.call_count == 3
+    assert mocked_lightkube_client.apply.call_count == 0
```
NohaIhab, line 62: same as `kfp-metadata-writer`

charms/kfp-api/src/templates/auth_manifests.yaml.j2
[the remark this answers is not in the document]
NohaIhab: As per @kimwnasptd's guidance, these CRDs are still in early stages so we should not include them for this release. Marking resolved.

charms/kfp-persistence/src/charm.py
```diff
@@ -33,7 +33,7 @@
 log = logging.getLogger()
 
 K8S_RESOURCE_FILES = ["src/templates/auth_manifests.yaml.j2"]
-SA_NAME = "ml-pipeline-persistenceagent"
+SA_NAME = "kfp-persistence"
```
orfeas-k, line 36: question: Why do we need this change?

charms/kfp-schedwf/src/components/sa_token_component.py
orfeas-k, line 1: Good job on figuring out this was already done in persistence agent and moving it here as well! I 'm wondering if that is a component that could live in [chisme](https://github.com/canonical/charmed-kubeflow-chisme/tree/main/src/charmed_kubeflow_chisme/components). Correct me if I'm wrong, it looks generic enough to me. If that's true, we can either:
1. create an issue to do this later
2. do it now and then update this PR to use the component from chisme

I prefer 2 tbh since 1 introduces tech debt. But if we 're really pressured by time, we can go with 1.

orfeas-k (commented):
Good job @MattiaSarti , the changes pretty much cover everything. I left two comments. WRT [envoy](https://github.com/canonical/envoy-operator), manifests update this image too but I do not see any [changes in upstream code](https://github.com/kubeflow/pipelines/tree/2.5.0/third_party/metadata_envoy) between the two tags, so we could skip this step, let's just mention it explicitly (in the PR/commit description).

charms/kfp-persistence/src/charm.py
orfeas-k: actually, is the token bound to this service account? I don't see any such service account being created anywhere, right? How was it working previously?

NohaIhab (approved):
+1, let's tackle the reusable component separately

Issue #781, opened 2025-09-11
Feature: Add KFP Launcher configuration functionality to the charm


toaksoy:
### Context

While troubleshooting https://github.com/canonical/charmed-kubeflow-uats/issues/210, we found out that if a different bucket name is chosen than the default one, the pipelines fail due to the fact that, even though the new bucket name is configured in the kfp-api charm, the charm doesn't create a kfp-launcher config to persist the new bucket name data.

https://www.kubeflow.org/docs/components/pipelines/operator-guides/configure-object-store/
https://www.kubeflow.org/docs/components/pipelines/concepts/pipeline-root/

### What needs to get done

Implement the functionality for the charm to create kfp-launcher config using the charm configurations, e.g. `object-store-bucket-name`

### Definition of Done

The charm ensures that the launcher config exists and is correct (according to charm configs) in the namespace, if it is not existent, then creates it foremost.

PR #633, opened 2024-12-06
chore: update python dependencies + pin pip

orfeas-k:
* Pin pip to 24.2 due to https://github.com/jazzband/pip-tools/issues/2131
* Update python dependencies using 'tox -e update-requirements'
* Remove unused `pytest-lazy-fixture` package

Ref canonical/bundle-kubeflow#1177
Ref canonical/bundle-kubeflow#1185

PR #787, opened 2025-09-26
feat: configure default pipeline root

MattiaSarti:
resolves https://github.com/canonical/kfp-operators/issues/781

Moreover, the opportunity is taken to address https://github.com/canonical/bundle-kubeflow/issues/1185 completely, and not just partially like in https://github.com/canonical/kfp-operators/pull/633, given that it was preventing the updated testing logic from being maintainable.

### DIFF
--- a/charms/kfp-profile-controller/tests/unit/test_operator.py
+++ b/charms/kfp-profile-controller/tests/unit/test_operator.py
@@ -29,7 +29,7 @@ CUSTOM_IMAGES_PATH = Path("./src/default-custom-images.json")
 with CUSTOM_IMAGES_PATH.open() as f:
     custom_images = json.load(f)
 
-KFP_DEFAULT_PIPELINE_ROOT = "minio://mlpipeline/v2/artifacts"
+KFP_DEFAULT_PIPELINE_ROOT = ""
 
 MOCK_OBJECT_STORAGE_DATA = {
     "access-key": "access-key",


charms/kfp-profile-controller/src/components/pebble_components.py
```diff
@@ -30,6 +30,12 @@ class KfpProfileControllerInputs:
 
 
 class KfpProfileControllerPebbleService(PebbleServiceComponent):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # NOTE: necessary for the (default) Pebble service to be updated on
+        # charm configuration changes, as its environment depends on them:
+        self._events_to_observe.append(getattr(self._charm.on, "config_changed"))
```
kimwnasptd, line 37: **question**: @misohu @NohaIhab does this mean that by default our charm's won't get triggered when we update a config value?

kimwnasptd: Aah, I see. @MattiaSarti I'd suggest let's have a bug that describes this current limitation of pebble component. So if in the future we bump into it again we can prioritise and add this functionality in the base component itself

misohu: Exactly as Matia described  ... wonder if we want to add it to every pebble comonent

PR #809, opened 2025-12-09
feat: run workload and charm as unprivileged user

### LINKED ISSUE #801: Add unprivileged configuration and tests for the KPF operators charm
### Context

As a necessary security measure, Charmed Kubeflow is supposed to run all charms and their respective workloads as non-root users in order to prevent threats that come with root privileges.

Assuming the [exploration](https://github.com/canonical/bundle-kubeflow/issues/1354) is completed, charm needs to be configured to run rootless and proper tests need to be implemented to ensure correct privileges.

### What needs to get done

1. Modify charm to run rootless
2. Add proper tests to ensure correct privileges

### Definition of Done

1. Charm is run rootless
2. Workload is run rootless
3. CI is passing

dariofaccin:
This PR introduces rootless platform-component workload and charms.
The workload container is run with user ID `584792` (which in the rock OCI image is mapped to user `_daemon_`).
The charm container is run with user `juju`, with static ID `170`.
The PR includes integration tests to assert that the security context of each container in the pod is correct.

### Removal of `user` parameter in `kfp-api`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

### Removal of `user` parameter in `kfp-metadata-writer`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

### Removal of `user` parameter in `kfp-persistence`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

### Update of default port in `kfp-profile-controller`
This PR changes the default binding port configuration for `kfp-profile-controller`, moving it to `1025`. This change is necessary since ports from `0` to `1024` require a privileged user. See [here](https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html) for more information.
Ths K8s service is exposed to previous port (`80`).

### Removal of `user` parameter in `kfp-schedwf`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

### Removal of `user` parameter in `kfp-ui`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

### Removal of `user` parameter in `kfp-viewer`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

### Removal of `user` parameter in `kfp-viz`
Since user is now enforced in `metadata.yaml`, its explicit set in the charm code is not required.

This PR closes #801.

### DIFF
--- a/charms/kfp-ui/src/charm.py
+++ b/charms/kfp-ui/src/charm.py
@@ -198,8 +198,10 @@ class KfpUiOperator(CharmBase):
         self.ml_pipeline_ui_container = self.charm_reconciler.add(
             component=MlPipelineUiPebbleService(
                 charm=self,
-                name=f"container:{self._container_name}",  # This feels a bit redundant, but will read
-                container_name=self._container_name,  # well in the statuses.  Thoughts?
+                # This feels a bit redundant, but will read
+                # well in the statuses.  Thoughts?
+                name=f"container:{self._container_name}",
+                container_name=self._container_name,
                 service_name="ml-pipeline-ui",
                 files_to_push=[
                     ContainerFileTemplate(


MattiaSarti (approved):
note: as for most charms, there are no other workloads in addition to the charm's ones, all already configured via respective `metadata.yaml`

Issue #843, opened 2026-01-30
kfp-schedwf: Hardcoded ServiceAccount name causes failure when deployed with custom application name


almeidaraul:
**Note: This issue was generated with AI assistance (GitHub Copilot) based on automated log analysis and triage.**
Filed by @canonical/solutions-qa

## Summary

The `kfp-schedwf` charm fails when deployed with an application name different from "kfp-schedwf" due to a hardcoded ServiceAccount name. The charm looks for a ServiceAccount named "kfp-schedwf" while Juju creates one matching the application name, resulting in HTTP 404 errors.

## Test Observer Link

https://test-observer.canonical.com/#/charms/320213?testExecutionId=319847&testResultId=9029691

## Details

When the charm is deployed as `target` (or any name other than "kfp-schedwf"), the following occurs:

1. Juju creates a Kubernetes ServiceAccount with the name `target` (matching the application name)
2. The charm attempts to create a token for ServiceAccount named `kfp-schedwf` (hardcoded value)
3. Kubernetes API returns 404: `serviceaccounts "kfp-schedwf" not found`
4. The charm enters a blocked state with the error: `Request to create token for kfp-schedwf ServiceAccount failed`

**Evidence from logs:**

Pod ServiceAccount (actual):
```
Service Account:  target
```

Error message (expected):
```
response body: {"kind":"Status","apiVersion":"v1","metadata":{},
"status":"Failure","message":"serviceaccounts \"kfp-schedwf\" not found",
"reason":"NotFound","details":{"name":"kfp-schedwf","kind":"serviceaccounts"},"code":404}
```

## Root Cause

The charm has a hardcoded ServiceAccount name at [line 38 of `charms/kfp-schedwf/src/charm.py`](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-schedwf/src/charm.py#L38):

```python
SA_NAME = "kfp-schedwf"
```

This constant is passed to the `SATokenComponent` at line 107-114, which then tries to access a ServiceAccount with this hardcoded name instead of using `self.app.name`.

**Fix needed:** Replace `SA_NAME = "kfp-schedwf"` with `SA_NAME = self.app.name` (or equivalent pattern used in the fix for kfp-persistence).

## Affected Test Executions

- Test Execution: 319847
- Test Result: 9029691
- Artefact: 320213 (kfp-schedwf revision 2505, channel 2.15/edge)
- Environment: kubernetes (k8s-production), ubuntu:24.04, amd64
[...]

PR #751, opened 2025-07-29
refactor: remove unnecessary roles, clusterroles and serviceaccounts

MattiaSarti:
resolves https://github.com/canonical/kfp-operators/issues/750

PR #844, opened 2026-01-30
Fix hardcoded ServiceAccount in kfp-schedwf

almeidaraul:
# Summary
This PR addresses the hardcoded ServiceAccount (SA) name in the `KfpSchedwf` class. Following the pattern established in #751 , the class now dynamically uses the app name as the ServiceAccount name.

# Changes
- **Logic update:** refactored `KfpSchedwf` to derive the ServiceAccount name from the application name rather than using a hardcoded string.
- **Testing:** added a new unit test to verify dynamic SA naming and updated an existing test that was previously dependent on the hardcoded "kfp-schedwf" name.

# Fixes
Fixes https://github.com/canonical/kfp-operators/issues/843

### DIFF
--- a/charms/kfp-schedwf/src/charm.py
+++ b/charms/kfp-schedwf/src/charm.py
@@ -49,7 +49,8 @@ class KfpSchedwf(CharmBase):
 
         self.charm_reconciler = CharmReconciler(self)
 
-        self.sa_name = self.app.name
+        self.app_name = self.app.name
+        self.sa_name = self.app_name
 
         self.leadership_gate = self.charm_reconciler.add(
             component=LeadershipGateComponent(
@@ -66,10 +67,10 @@ class KfpSchedwf(CharmBase):
                 resource_templates=K8S_RESOURCE_FILES,
                 krh_resource_types={CustomResourceDefinition},
                 krh_labels=create_charm_default_labels(
-                    self.app.name, self.model.name, scope="crds"
+                    self.app_name, self.model.name, scope="crds"
                 ),
                 context_callable=lambda: {
-                    "app_name": self.app.name,
+                    "app_name": self.app_name,
                     "namespace": self.model.name,
                     "sa_name": self.sa_name,
                 },


charms/kfp-schedwf/lib/charms/istio_beacon_k8s/v0/service_mesh.py
MattiaSarti, line 1: suggestion (blocking): the source of the charm libraries is [auto-generated](https://documentation.ubuntu.com/charmcraft/stable/reference/commands/fetch-libs/#fetch-libs) and we would rather exclude them from the linting process than change their source (even if just to fix their style), like [we already do](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-schedwf/tox.ini#L51), so please feel free to restore this file

charms/kfp-schedwf/src/charm.py
```diff
@@ -72,7 +73,7 @@ def __init__(self, *args):
                 context_callable=lambda: {
                     "app_name": self.app.name,
```
MattiaSarti: suggestion (non-blocking): can we also define a common attribute - like `self.app_name` - in order not to repeat `self.app.name` above and across these lines here, and then assign it this way `self.sa_name = self.app_name` to make the fact we choose the app name as the ServiceAccount name explicit while avoiding repetitions?

deusebio (commented):
Changes looks good! Could you just rebase such that we get green CI?

Once this is done, I'll approve and merge the PR. 

Thanks!

almeidaraul:
@MattiaSarti would you mind verifying that the remaining CI issues aren't related to these changes?

MattiaSarti:
> @MattiaSarti would you mind verifying that the remaining CI issues aren't related to these changes?

Sure, @almeidaraul.

Some test failures were due to some "flakiness" of our CI, so I've rerun the relevant ones and they all eventually passed. The remaining failures are still unrelated to your changes, to me, and those tests are less relevant. So I would consider your changes safe.

PR #856, opened 2026-02-10
[KF-8438] (kfp-profile-controller) Remove statefulness from charm con…

deusebio:
…figuration

This requires the new python image to be released, as raise inthe PR: https://github.com/canonical/python-rock/pull/37

### DIFF
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -115,6 +115,8 @@ class KfpProfileControllerOperator(CharmBase):
             return
         self.default_pipeline_root = self.model.config["default_pipeline_root"]
 
+        self._container_name = next(iter(self.meta.containers))
+
         # expose controller's port
         http_port = ServicePort(K8S_SVC_CONTROLLER_PORT, name="http", targetPort=CONTROLLER_PORT)
         self.service_patcher = KubernetesServicePatch(
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -64,6 +64,8 @@ METADATA_GRPC_SERVICE_PORT = "8080"
 NAMESPACE_LABEL = "pipelines.kubeflow.org/enabled"
 SYNC_CODE_FILE = Path("files/upstream/sync.py")
 
+HOOKS_PATH = Path("/run")
+
 
 def parse_images_config(config: str) -> Dict:
     """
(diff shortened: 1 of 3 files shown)


charms/kfp-profile-controller/src/components/pebble_components.py
```diff
@@ -59,7 +60,7 @@ def get_layer(self) -> Layer:
                     self.service_name: {
                         "override": "replace",
                         "summary": "entry point for kfp-profile-controller",
-                        "command": "python3 /hooks/sync.py",  # Must be a string
+                        "command": f"python3 {str(inputs.HOOKS_PATH / 'sync.py')}",
```
dariofaccin, line 64: Why are we casting to `str` if the attribute is already defined as `str` in L32?

charms/kfp-profile-controller/src/charm.py
```diff
@@ -221,6 +219,7 @@ def __init__(self, *args):
                     FRONTEND_TAG=self.images["frontend__version"],
                     KFP_API_PRINCIPAL=str(self.model.config["kfp-api-principal"]),
                     AMBIENT_ENABLED=self.service_mesh_component.component.ambient_mesh_enabled,
+                    HOOKS_PATH=HOOKS_PATH,
```
dariofaccin, line 222: If `HOOKS_PATH` attribute is `str`, we probably want to cast it here.

dariofaccin (changes requested):

charms/kfp-profile-controller/src/components/pebble_components.py
MattiaSarti: I think it's a typo with the type hint in line 32, because it actually gets a `Path` instance from `charm.py` and casting seems necessary

charms/kfp-profile-controller/src/charm.py
MattiaSarti: +1

MattiaSarti (approved):

MattiaSarti (approved):

charms/kfp-profile-controller/src/components/pebble_components.py
deusebio: I changed the type of the input, such that I can use the nice syntactic sugar `Path / <file>`

charms/kfp-profile-controller/src/charm.py
deusebio: I changed the type of "HOOKS_PATH" into Path

deusebio:
Folks, it [turns out](https://github.com/canonical/python-rock/pull/37#issuecomment-3885493686) that there is a kinda hidden home in the python image, under `/var/lib/pebble/default`. This path is indeed writable. I have update the PR accordingly, while reviewing and addressing your comments

dariofaccin (approved):

PR #862, opened 2026-02-13
[MISC] ci: ensuring to apply release revision

deusebio:
This ensures that when testing tf apply, we consume from the channel that we have released to. This therefore does not make it necessary to change the default channel in TF modules that has not yet been released to stable

[Reference of the conversation](https://chat.canonical.com/canonical/pl/8w4anegip7r5je9rfpff6wf8zr)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -109,8 +109,10 @@ jobs:
 
           if "${{ github.event_name }}" == "pull_request":
               ref = "${{ github.base_ref }}"
+              risk = "pr-${{ github.event.pull_request.number }}"
           else:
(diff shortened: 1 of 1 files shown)


misohu (approved):

deusebio:
I had to open a couple of PR, since initially I opened a PR against the feature branch and that was releasing to `latest/edge/pr-XXX`. I then re-opened the PR against the `track` to make sure that the charm was released to the right track and referenced, as you can see [here](https://github.com/canonical/kfp-operators/actions/runs/22060332823/job/63739566645?pr=864#step:6:63)

mvlassis (approved):

PR #868, opened 2026-03-24
feat: add automated backport support

mvlassis:
Base issue: https://github.com/canonical/bundle-kubeflow/issues/1421

This PR enables automatic creation of backports based on labels. Labels are automatically assigned on each new PR based on the value of the `.github/automatic_backport_tracks`.

### DIFF
--- a/.github/automatic_backport_tracks.yaml
+++ b/.github/automatic_backport_tracks.yaml
@@ -1,0 +1,2 @@
+automatic_backport_tracks:
+  -   track/latest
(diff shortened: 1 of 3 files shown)


.github/automatic_backport_tracks.yaml
```diff
@@ -0,0 +1,2 @@
+automatic_backport_tracks:
+  -   track/2.15
```
MattiaSarti, line 2: issue: how about tracks for CKF 1.10 (here `2.5`)? Should we keep backporting to them as well?

dariofaccin (approved):

dariofaccin (changes requested):
suggestion(blocking): we should support both `track/2.5` and `track/2.15`

MattiaSarti (approved):

dariofaccin (approved):

Issue #869, opened 2026-03-30
Base task to upgrade manifests for pipelines to `2.16`


dariofaccin:
### Context

A [new version](https://github.com/kubeflow/pipelines/tree/2.16.0) of Kubeflow Pipelines is available. The scope of this task is to upgrade the manifests of all charms to reflect new versions.

### What needs to get done

1. Upgrade manifests for `kfp-api`
2. Upgrade manifests for `kfp-metadata-writer`
3. Upgrade manifests for `kfp-persistence`
4. Upgrade manifests for `kfp-profile-controller`
5. Upgrade manifests for `kfp-schedwf`
6. Upgrade manifests for `kfp-ui`
7. Upgrade manifests for `kfp-viewer`
8. Upgrade manifests for `kfp-viz`

### Definition of Done

1. Manifests are upgraded

mvlassis:
Closed by https://github.com/canonical/kfp-operators/pull/870

Issue #822, opened 2026-01-16
Update logic in `sync.py` to match the effort in upstream


mvlassis:
### Context
In https://github.com/kubeflow/manifests/pull/3283, `sync.py` was updated to work with the change to `seaweedfs` for S3 storage. 

The main change was to allow multi-user isolation in pipelines, as pipelines artifacts are now stored in separate buckets instead of the one at `KFP_DEFAULT_PIPELINE_ROOT`. The update also introduced changes to the creation of `Deployments`, as it doesn't create `Deployments` for `ml-pipeline-visualizationserver` and `ml-pipeline-ui-artifact` in the user namespace.

The diff between the 2 changes can be found [here](https://www.diffchecker.com/HXIGkIpW/).

We should investigate how we want to modify the file to fit with our plan for S3 storage.

mvlassis:
## Exploration

### Handling of `ml-pipeline-ui` deployments in the user namespace
- `sync.py` was updated  in https://github.com/kubeflow/pipelines/pull/11965 to conditionally create the `ml-pipeline-ui` deployments in the user namespace as well as the respective service, see https://github.com/kubeflow/pipelines/blob/bc89227f26afb8c5601581c2b7b5634ce017728d/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L230.
- This condition is based on the value of the environment variable `ARTIFACTS_PROXY_ENABLED`.
- At the same time, the function [getArtifactsProxyHandler](https://github.com/kubeflow/pipelines/blob/bc89227f26afb8c5601581c2b7b5634ce017728d/frontend/server/handlers/artifacts.ts#L437) used for the `frontend` image in the `kfp-ui` conditionally uses this environment variable to redirect requests to the `ml-pipeline-ui` service in the user deployment.
- The default value of `ARTIFACTS_PROXY_ENABLED` was changed to `false` in https://github.com/kubeflow/pipelines/pull/11965. Comments in https://github.com/kubeflow/manifests/blob/1a12e1be42f69ad4a3d374dc3e9cb2da6eb26706/applications/pipeline/upstream/base/installs/generic/pipeline-install-config.yaml#L97-L104 explain that this change was made for security reasons, the reasoning behind the change being still unclear.

## Miscellaneous
- Change image of `kfp-profile-controller` from the current Python one to `alpine/k8s`, see [this PR](https://github.com/kubeflow/pipelines/pull/11965/changes).

PR #870, opened 2026-04-02
feat: Update manifests to match upstream kfp `2.16.0` release

mvlassis:
Ref: #869

This PR updates the charms according to the update to the manifests files corresponding to the `2.16.0` release of Kubeflow pipelines. This task is part of the team's upgrade to Kubeflow `26.03`.

The diff between the 2 versions was generated by following the instructions on the [upstream README](https://github.com/kubeflow/manifests?tab=readme-ov-file#pipeline-definitions-stored-as-kubernetes-resources). It can be found [here](https://www.diffchecker.com/NW7YlN8J/).

## Notes
- Update `kfp` rocks to the newer upstream versions.
- Replace `envoy` rock with `ghcr.io/kubeflow/kfp-metadata-envoy:2.16.0`, since it is now using [v1.37.0](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/third_party/metadata_envoy/Dockerfile#L15). See https://github.com/canonical/envoy-operator/pull/203.
- Update deployment for `mlmd`, see https://github.com/canonical/mlmd-operator/pull/173.
- A lot of deployments were updated to set a `seccompProfile`, which is a security setting for limiting syscalls within the containers. For more details, see the [Kubernetes documentation](https://kubernetes.io/docs/tutorials/security/seccomp/). Unfortunately, this is *not* possible to use when using Juju for deploying the workloads.
- In a similar manner, we *cannot* specifically drop all Linux capabilities, which is not used in the `workflow-controller-configmap` configMap. 
- In parallel, some deployments added a `runAsNonRoot: true` key-value pair, which we already implement in our charms by setting the `uid` and `gid` to a non-root value in the `charmcraft.yaml` files.
- `kfp-schedwf` has introduced metrics in https://github.com/kubeflow/pipelines/pull/12610. I have created https://github.com/canonical/kfp-operators/issues/871 to track the respective effort in CKF.
- The addition of the `kubeflow_pipelines_version` value in a new configMap seems redundant, as it is only used in the repository for denoting new versions, and never in the actual code. See the [Pipelines README](https://github.com/kubeflow/pipelines/blob/e7d6aa73d69cc6690105d2e622409fe0a22e97ba/RELEASE.md?plain=1#L157).
- A few changes (e.g. adding `OBJECT_STORE_HOST`, `CLUSTER_DOMAIN`, changes to `sync.py`) correspond to the change from `minio` to using `seaweedfs` with separate user buckets. We do not include those changes here, the changes are tracked in https://github.com/canonical/kfp-operators/issues/822
- Remove comments for updating the user when changing the OCI from upstream to a rock and vice-versa. This is because we don't anymore use a `service-config.yaml`.


charms/kfp-api/config.yaml
MattiaSarti, line 1: suggestion (blocking): despite reading your notes:

> A few changes (e.g. adding OBJECT_STORE_HOST, CLUSTER_DOMAIN, changes to sync.py) correspond to the change from minio to using seaweedfs with separate user buckets. We do not include those changes here, the changes are tracked in https://github.com/canonical/kfp-operators/issues/822

> In parallel, some deployments added a runAsNonRoot: true key-value pair, which we already implement in our charms by setting the uid and gid to a non-root value in the charmcraft.yaml files.

we still need to make sure all new environment variables are passed, because the inner image logic may otherwise break, even if they are not used without Seaweed or if they don't end up changing any `securityContext`, so I would still add the following ones to `kfp.api`, :
- `CLUSTER_DOMAIN` (also, see [this comment](https://github.com/canonical/kfp-operators/pull/870/changes#r3030818629))
- `DEFAULT_SECURITY_CONTEXT_RUN_AS_USER`
- `DEFAULT_SECURITY_CONTEXT_RUN_AS_GROUP`
- `DEFAULT_SECURITY_CONTEXT_RUN_AS_NON_ROOT`

charms/kfp-schedwf/src/components/pebble_component.py
MattiaSarti, line 1: suggestion (blocking): for the same reasons as the comment above, despite your note:

> kfp-schedwf has introduced metrics in https://github.com/kubeflow/pipelines/pull/12610. I have created https://github.com/canonical/kfp-operators/issues/871 to track the respective effort in CKF.

I would still add the environment variable `METRICS_PORT`, even if unused

MattiaSarti, line 1: question: is the container port change to add `9090` excluded because considered as part of this?

> kfp-schedwf has introduced metrics in https://github.com/kubeflow/pipelines/pull/12610. I have created https://github.com/canonical/kfp-operators/issues/871 to track the respective effort in CKF.

charms/kfp-ui/src/components/pebble_components.py
MattiaSarti, line 1: issue: we said that:

> A few changes (e.g. adding OBJECT_STORE_HOST, CLUSTER_DOMAIN, changes to sync.py) correspond to the change from minio to using seaweedfs with separate user buckets. We do not include those changes here, the changes are tracked in https://github.com/canonical/kfp-operators/issues/822

yet we include changes to `CLUSTER_DOMAIN`

Nevertheless, I would leave these changes, and I have also suggested adding it [here](https://github.com/canonical/kfp-operators/pull/870/changes#r3030790497) coherently

MattiaSarti (changes requested):

charms/kfp-schedwf/src/components/pebble_component.py
mvlassis: Yes, that should be part of https://github.com/canonical/kfp-operators/issues/871

mvlassis: Added in https://github.com/canonical/kfp-operators/pull/870/commits/61aceb32b287aad32ef109569fea7b2f2798939d

charms/kfp-api/config.yaml
mvlassis: I don't think that these env variables missing would break the deployment:
- `CLUSTER_DOMAIN` uses the default values that Kubernetes already uses, see [here](https://github.com/kubeflow/pipelines/blob/e7d6aa73d69cc6690105d2e622409fe0a22e97ba/backend/metadata_writer/src/metadata_helpers.py#L36)
- The `DEFAULT_SECURITY_CONTEXT` ones already have empty values by default, see [here](https://github.com/kubeflow/pipelines/blob/e7d6aa73d69cc6690105d2e622409fe0a22e97ba/backend/src/apiserver/common/config.go#L179-L189)

That said, to more align with the upstream script, I added them in https://github.com/canonical/kfp-operators/pull/870/commits/1a060e3bcd4ac02ce98d7afa735f16d0b65bacd9

MattiaSarti: Thanks for linking those references! It's that just, considering their images as black boxes, you never know if empty variables are still read and the image logic breaks if they are not defined or if the logic handles that by assuming them empty when not defined.

charms/kfp-ui/src/components/pebble_components.py
MattiaSarti: question: @mvlassis, does [the code reference shared above](https://github.com/kubeflow/pipelines/blob/e7d6aa73d69cc6690105d2e622409fe0a22e97ba/backend/metadata_writer/src/metadata_helpers.py#L36) for `CLUSTER_DOMAIN` in response to my comment about `kfp-api` apply to `kfp-ui` as well?

MattiaSarti: Sorry, I still miss this, it's just for me to understand

MattiaSarti (changes requested):
suggestion (blocking): for the charm workloads run as root user, I have [the same objection I raised in another similar pull request](https://github.com/canonical/envoy-operator/pull/203#pullrequestreview-4062008653)

mvlassis:
Regarding the bundle tests, there is an issue with the bucket location which I'm trying to debug.

mvlassis:
## Note - new component
As explained in https://github.com/canonical/kfp-operators/issues/872#issuecomment-4281947359:

[This change](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/backend/src/apiserver/client_manager/client_manager.go#L958) was introduced to not exit when the bucket can not be created. We can see from the logs, that even though the bucket is not created, the [initBlobObjectStore](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/backend/src/apiserver/client_manager/client_manager.go#L948) function exits successfully:

```
kk logs kfp-api-0 -c apiserver | grep -i "succes"
2026-04-20T14:57:46.021Z [apiserver] I0420 14:57:46.020945      18 client_manager.go:279] DB client initialized successfully
2026-04-20T14:57:46.810Z [apiserver] I0420 14:57:46.810711      18 client_manager.go:1144] Successfully created bucket mlpipeline
2026-04-20T14:57:46.810Z [apiserver] I0420 14:57:46.810828      18 client_manager.go:967] Successfully initialized blob storage for bucket: mlpipeline
2026-04-20T14:57:46.810Z [apiserver] I0420 14:57:46.810967      18 client_manager.go:311] Object store client initialized successfully
2026-04-20T14:57:46.811Z [apiserver] I0420 14:57:46.811675      18 client_manager.go:324] Client manager initialized successfully
2026-04-20T14:57:46.905Z [apiserver] I0420 14:57:46.905406      18 config.go:154] Successfully uploaded Pipeline tutorial-data-passing-in-python-components.
2026-04-20T14:57:47.047Z [apiserver] I0420 14:57:47.047683      18 config.go:206] Successfully uploaded PipelineVersion tutorial-data-passing-in-python-components.
2026-04-20T14:57:47.100Z [apiserver] I0420 14:57:47.100426      18 config.go:154] Successfully uploaded Pipeline tutorial-dsl-control-structures.
2026-04-20T14:57:47.220Z [apiserver] I0420 14:57:47.220188      18 config.go:206] Successfully uploaded PipelineVersion tutorial-dsl-control-structures.
```

Thus, the Pebble service never restarts, and the executable never attempts to create the bucket again, unless we force a Pebble restart (for example, via reconfiguring minio).

#### Why initBlobObjectStore exists successfully
This functions calls [openBucketWithRetry](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/backend/src/apiserver/client_manager/client_manager.go#L1084), which in turns calls [OpenBucketV2](https://github.com/google/go-cloud/blob/78b59762fe5eb36407d2462399954d0d6e88ada9/blob/s3blob/s3blob.go#L244). The issue is that this function only fails when insufficient options have been provided, and doesn't actually check that the bucket actually exists in the specified location. Thus, this seems like a false positive where the apiserver incorrectly assumes that the bucket exists, while in reality it doesn't.


### Solution on charm.py
Since `apiserver` needs to be able to create a Bucket on the object storage, we have added a new method on `charm.py` named `_check_object_storage_reachable`, that tries to create a sample bucket `kfp-api-healthcheck` on the object storage. This needs to succeed before the respective Pebble component for `apiserver` executes. 

To enable this, we have also updated the respective unit tests with new mocks.

charms/kfp-api/src/charm.py
```diff
@@ -831,6 +836,48 @@ def _on_event(self, event, force_conflicts: bool = False) -> None:
 
         self.model.unit.status = ActiveStatus()
 
+    def _check_object_storage_reachable(self) -> None:
+        """Ensure object storage is reachable using a boto3 client."""
+        try:
```
misohu: suggestion (blocking): I like the solution but: 
- maybe i would move it to separate module like here https://github.com/canonical/mlflow-operator/blob/main/src/services/s3.py#L11
- also those catch all exceptions are not a good idea we should be catching only expected exceptions ... also dont do string matching for errors use rather proper boto errors objects
- you should also remove the bucket at the end 
- I would also split this to smaller testable helper functions e.g.:
build_s3_client
create_probe_bucket
delete_probe_bucket
check_object_storage_reachable


**Note:** THis thing can put the charm to sleep until update status. I think there is nothing better we can do about it ?

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -175,19 +175,39 @@ def sync(self, parent, attachments):
 
             desired_configmap_count = 1
             desired_resources = []
+
+            providers_yaml = (
+                "s3:\n"
+                "  default:\n"
+                f"    endpoint: {minio_host}.{minio_namespace}:9000\n"
+                "    disableSSL: true\n"
+                "    region: us-east-1\n"
```
misohu, line 194: suggestion (blocking): should this be hard coded?

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -175,19 +175,39 @@ def sync(self, parent, attachments):
...
+                "    forcePathStyle: true\n"
+                "    credentials:\n"
+                "      fromEnv: false\n"
+                "      secretRef:\n"
+                "        secretName: mlpipeline-minio-artifact\n"
```
misohu, line 199: suggestion (blocking): should this be hard coded?

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -175,19 +175,39 @@ def sync(self, parent, attachments):
...
-                    },
-                }]
+                configmap_data["defaultPipelineRoot"] = kfp_default_pipeline_root
+
+            desired_configmap_count = 2
+            desired_resources += [{
+                "apiVersion": "v1",
+                "kind": "ConfigMap",
+                "metadata": {
+                    "name": "kfp-launcher",
```
misohu, line 215: suggestion (blocking): should this be hard coded? if so at least have a constant

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -175,19 +175,39 @@ def sync(self, parent, attachments):
...
-                        "name": "kfp-launcher",
-                        "namespace": namespace,
-                    },
-                    "data": {
-                        "defaultPipelineRoot": kfp_default_pipeline_root,
-                    },
-                }]
+                configmap_data["defaultPipelineRoot"] = kfp_default_pipeline_root
+
+            desired_configmap_count = 2
```
misohu: suggestion (blocking): another suspicious magic number... it can be a problem in future changes

tests/integration/test_kfp_functional.py
```diff
@@ -146,8 +166,11 @@ def test_upload_pipeline(kfp_client):
     kfp_client.delete_pipeline(pipeline_upload_response.pipeline_id)
 
 
-def test_create_and_monitor_run(kfp_client, create_and_clean_experiment_v2):
+def test_create_and_monitor_run(lightkube_client, kfp_client, create_and_clean_experiment_v2):
     """Create a run and monitor it to completion."""
+    # Ensure "kfp-launcher" configmap has been created on the profile namespace
+    wait_for_configmap(lightkube_client, KFP_LAUNCHER_CONFIGMAP_NAME, KUBEFLOW_PROFILE_NAMESPACE)
```
misohu: suggestion(blocking): why we do this on every test ? we can do this once when we create the namespace in the fixture

tests/integration/test_kfp_functional_ambient.py
```diff
@@ -148,8 +168,11 @@ def test_upload_pipeline(kfp_client):
     kfp_client.delete_pipeline(pipeline_upload_response.pipeline_id)
 
 
-def test_create_and_monitor_run(kfp_client, create_and_clean_experiment_v2):
+def test_create_and_monitor_run(lightkube_client, kfp_client, create_and_clean_experiment_v2):
     """Create a run and monitor it to completion."""
+    # Ensure "kfp-launcher" configmap has been created on the profile namespace
+    wait_for_configmap(lightkube_client, KFP_LAUNCHER_CONFIGMAP_NAME, KUBEFLOW_PROFILE_NAMESPACE)
```
misohu: Same comment as in previous file

charms/kfp-profile-controller/files/upstream/sync.py
mvlassis: Here we follow the same structure as [upstream](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L203) which hardcodes it. Since we want to follow the upstream `sync.py` as close as possible, I think it's OK in this case

charms/kfp-profile-controller/files/upstream/sync.py
mvlassis: Same as above, we follow the [upstream structure](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L191). I updated `sync.py` in https://github.com/canonical/kfp-operators/pull/870/commits/b3d8b2d2cd62398a0f6aff2477b2f0599b9406e0 to be more clear on which configMaps should created above the definition of `desired_configmap_count`.

charms/kfp-profile-controller/files/upstream/sync.py
mvlassis: The secret name is already hardcoded in `sync.py` (see [here](https://github.com/canonical/kfp-operators/blob/c5f14c14dd98ec969578fe32efcff343fe004802/charms/kfp-profile-controller/files/upstream/sync.py#L523-L535)), and the same behavior is [upstream](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L419).

tests/integration/test_kfp_functional_ambient.py
mvlassis: Moved under the `apply_profile` fixture in `confest.py` in https://github.com/canonical/kfp-operators/pull/870/commits/581be2b6013b481ad89d049f0dfc419d5862a8f9

tests/integration/test_kfp_functional.py
mvlassis: Moved under the `apply_profile` fixture in `confest.py` in https://github.com/canonical/kfp-operators/pull/870/commits/581be2b6013b481ad89d049f0dfc419d5862a8f9

charms/kfp-profile-controller/files/upstream/sync.py
mvlassis: ### Is region mandatory?
In `2.16.0`, the field itself is mandatory in the context of the `apiserver`. Removing this field produces an error:
```
k logs -n kubeflow-user-example-com v2-container-component-no-input-5vsn6-system-container-impl-1485897518
time="2026-04-23T11:35:41.390Z" level=info msg="capturing logs" argo=true
I0423 11:35:41.475714      16 main.go:66] Setting log level to: '1'
I0423 11:35:41.481338      16 cache.go:89] Connecting to cache endpoint ml-pipeline.kubeflow.svc.cluster.local:8887
hello world
I0423 11:35:41.545904      16 launcher_v2.go:1058] output metadata file does not exist in /s3/mlpipeline/v2/artifacts/v2-container-component-no-input/1ded32bd-2940-4b79-85ef-693d4e5cc1d8/container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/output_metadata.json
I0423 11:35:41.545945      16 launcher_v2.go:695] outputArtifact in uploadOutputArtifacts call: executor-logs
E0423 11:35:41.549795      16 launcher_v2.go:445] Failed to upload output artifacts: failed to upload output artifact "executor-logs" to remote storage URI "s3://mlpipeline/v2/artifacts/v2-container-component-no-input/1ded32bd-2940-4b79-85ef-693d4e5cc1d8/container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs": uploadFile(): unable to complete copying "/s3/mlpipeline/v2/artifacts/v2-container-component-no-input/1ded32bd-2940-4b79-85ef-693d4e5cc1d8/container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs" to remote storage "container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs": failed to close Writer for bucket: blob (key "container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs") (code=Unknown): operation error S3: PutObject, failed to resolve service endpoint, endpoint rule error, A region must be set when sending requests to S3.
I0423 11:35:41.549842      16 launcher_v2.go:447] Refreshing credentials before retrying artifacts upload.
I0423 11:35:41.551905      16 launcher_v2.go:458] Executing second uploadOutputArtifacts attempt.
I0423 11:35:41.551922      16 launcher_v2.go:695] outputArtifact in uploadOutputArtifacts call: executor-logs
I0423 11:35:41.587987      16 launcher_v2.go:201] publish success.
I0423 11:35:41.598919      16 client.go:777] Attempting to update DAG state
F0423 11:35:41.626910      16 main.go:58] failed to execute component: failed to upload output artifact "executor-logs" to remote storage URI "s3://mlpipeline/v2/artifacts/v2-container-component-no-input/1ded32bd-2940-4b79-85ef-693d4e5cc1d8/container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs": uploadFile(): unable to complete copying "/s3/mlpipeline/v2/artifacts/v2-container-component-no-input/1ded32bd-2940-4b79-85ef-693d4e5cc1d8/container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs" to remote storage "container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs": failed to close Writer for bucket: blob (key "container-no-input/da6062d6-9bd0-4b15-b8b5-9c34c1cf9202/executor-logs") (code=Unknown): operation error S3: PutObject, failed to resolve service endpoint, endpoint rule error, A region must be set when sending requests to S3.
time="2026-04-23T11:35:42.391Z" level=info msg="sub-process exited" argo=true error="<nil>"
Error: exit status 1
```

### Behavior with non-AWS storage
For non-AWS S3 storage (which includes the `minio` we use in our bundle), this value is [optional](https://github.com/kubeflow/pipelines/blob/24ce44baf0070f76dcaf25d91a8c111de065c73f/backend/src/v2/config/s3.go#L34) and thus not used at all. We can use any value, even `minio`, but `us-east-1` makes more sense to me since this is a valid AWS region.

### Behavior with AWS storage
For AWS S3 storage, this region is used in the [S3Config](https://github.com/kubeflow/pipelines/blob/e4ebca310f404dac306e16bc880de12bbbf63b95/backend/src/v2/objectstore/object_store.go#L269). This could affect us when using `minio-operator` in [gateway mode](https://documentation.ubuntu.com/charmed-kubeflow/latest/how-to/manage/integrate-minio/). 

Currently in `gateway-mode`, we don't specify a region in `minio-operator`, and users are asked to set the region in the `endpoint` config option. This endpoint is passed to `kfp-api` and a client to connect with AWS is created. What's not clear here is how the `region` is populated in `gateway-mode`.

charms/kfp-api/src/charm.py
mvlassis: #### Proposal - Running only once
An idea would be to only run `check_object_storage_reachable` until it succeeds once. We can use a simple flag for this, initialized to `True`. After the check succeeds, we would mark the flag as `False`, and start `apiserver`. We should mark the flag as `True` whenever the endpoint changes, or whenever the relation is broken.

The issue with this proposal is that it would break the contract for the "Reconciler" pattern we use in this charm. We already run everything on `_on_event`, including applying manifest files, and the proposal would require saving the current state of the flag in permanent storage, and the method would only be run depending on this state.

A good compromise here would be to use a specific timeout to not wait for a long time. I updated https://github.com/canonical/kfp-operators/pull/870/commits/30b0fffc50069ad61c03834e90dba288aac82650 to not hang for minutes if a connection fails.

mvlassis:
After discussing with @misohu, we:
- Try to create the same bucket that `apiserver` needs, in order to avoid creating unnecessary buckets.
- ~Defer the event in case the object storage is not reachable. This follows the [good practices](https://documentation.ubuntu.com/ops/latest/explanation/defer-guidance/#good-retrying-on-temporary-failure) for deferring an event, since the error should only be temporary (during startup of `minio`).~ We prefer not to defer on reconciler patterns, so we just wait until the next event is triggered to put the status to "active".

misohu (approved):

misohu (commented):
Wait we never defer in our charms we use update status

charms/kfp-api/src/charm.py
NohaIhab, line 1: question: what about the `kubeflow-pipelines-public` that's been added upstream? Just came across it in diffchecker

charms/kfp-api/src/charm.py
```diff
@@ -409,6 +412,9 @@ def _generate_environment(self) -> dict:
             "OBJECTSTORECONFIG_HOST": f"{object_storage['service']}.{object_storage['namespace']}",
             "OBJECTSTORECONFIG_PORT": str(object_storage["port"]),
             "OBJECTSTORECONFIG_REGION": "",
+            "DEFAULT_SECURITY_CONTEXT_RUN_AS_USER": "",
+            "DEFAULT_SECURITY_CONTEXT_RUN_AS_GROUP": "",
+            "DEFAULT_SECURITY_CONTEXT_RUN_AS_NON_ROOT": "",
```
NohaIhab: what does it mean to set those to empty string? can you add a comment?

charms/kfp-api/poetry.lock
NohaIhab, line 1: suggestion(non-blocking): should we also re-generate the root poetry.lock to update the sdk version?

charms/kfp-api/metadata.yaml
```diff
@@ -11,14 +11,13 @@ issues: https://github.com/canonical/kfp-operators/issues
...
     resource: oci-image
-    uid: 584792
-    gid: 584792
+    uid: 0
+    gid: 0
 resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
-    # The container's `user` needs to be updated when switching from upstream image to rock
```
NohaIhab, line 20: nit: keep the comment

NohaIhab: or rather add one like the other metadata.yaml:
```
# Set these to 584792 when using the rock
```

charms/kfp-profile-controller/tests/integration/test_charm.py
```diff
@@ -330,6 +340,13 @@ async def test_first_change_to_config_for_deafult_pipeline_root(
         apps=[CHARM_NAME], status="active", raise_on_blocked=True, timeout=300
     )
 
+    # NOTE: simulating the necessary manual deletion of the old ConfigMap by the user:
+    # https://github.com/kubeflow/manifests/blob/v1.11.0/applications/pipeline/upstream/base/installs/generic/pipeline-install-config.yaml#L40-L42  # noqa: E501 # fmt: skip
```
NohaIhab, line 344: praise: helpful comment

charms/kfp-ui/src/components/pebble_components.py
mvlassis: Sorry for replying late, yes! It should be defined [here](https://github.com/kubeflow/pipelines/blob/23fa62b361e5834694be8a74b536ce5035b2ab4e/backend/src/apiserver/common/const.go#L82), with the default value being "cluster.local". For completeness I added it it https://github.com/canonical/kfp-operators/pull/870/commits/3b5f7686a86cd18370305c44c175641d694b1d4d

charms/kfp-profile-controller/files/upstream/sync.py
```diff
@@ -173,21 +173,41 @@ def sync(self, parent, attachments):
                 logger.info(f"Namespace not in scope, no action taken (metadata.labels.pipelines.kubeflow.org/enabled = {pipeline_enabled}, must be 'true')")
                 return {"status": {}, "attachments": []}
 
-            desired_configmap_count = 1
+            # metadata-grpc-configmap and kfp-launcher are always created
```
NohaIhab: suggestion(blocking): can we explain that this is (yet another) deviation from upstream in a comment and why we need it? You put a nice comment in the PR but this will not be remembered when making code changes

NohaIhab (commented):
Thanks @mvlassis, some comments and questions for me to understand

charms/kfp-ui/src/components/pebble_components.py
MattiaSarti: No problem! Thanks!

charms/kfp-api/metadata.yaml
mvlassis: Added in https://github.com/canonical/kfp-operators/pull/870/commits/2857cee6bbc32f8659b8cfb7dbfb63f3c84dbbb8

charms/kfp-api/poetry.lock
mvlassis: That's a great catch! Updated in https://github.com/canonical/kfp-operators/pull/870/commits/3852410cd050daf3a847847832805657bc11b7de

charms/kfp-profile-controller/files/upstream/sync.py
mvlassis: Added in https://github.com/canonical/kfp-operators/pull/870/commits/cd48b1fa28dfd15635854520c8793ccc9bbbe55b

charms/kfp-api/src/charm.py
mvlassis: ### Context
Upstream added [3 options](https://github.com/kubeflow/manifests/blob/c4d60708c619856cca8020853eeb8cc76dbf2f75/applications/pipeline/upstream/base/installs/generic/pipeline-install-config.yaml#L114-L116) to `pipeline-install-config` which is the configMap they use to configure the `apiserver`. The options are:
- `runAsUser`
- `runAsGroup`
- `runAsNonRoot`

What they essentially do it (if set) is create a `securityContext` for the `main` container in a pipeline pod. An example would be:
```yaml
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsGroup: 0
      runAsNonRoot: false
      runAsUser: 1001
      seccompProfile:
        type: RuntimeDefault
```
If these are not set, then the respective field is not populated. This is to respect backward compatibility, because for example setting `runAsNonRoot` to `true` would disallow any container running as root. 

These 3 configuration options are meant to be for the cluster administrator, as explained in https://github.com/kubeflow/pipelines/pull/12859. `runAsUser` and `runAsGroup` don't seem very useful, but `runAsNonRoot` can be useful for admins who want to enforce a stricter setting on what can be run in pipelines.

### Decision
Being able to configure these options is more of a product decision. My intuition is to add them as we do for other values in `pipelines-install-config` like the log level. Added them in https://github.com/canonical/kfp-operators/pull/870/commits/6f4695f91dc2a628a1d8b35c7330e38b56f2aaa3

charms/kfp-api/src/charm.py
mvlassis: This change was introduced in https://github.com/kubeflow/pipelines/pull/12717.

This PR includes the following:
1. Add a `kubeflow-pipelines-public` configMap with a single key:value: `kubeflow_pipelines_version=v2.16.0`, with the value corresponding to the current version of the pipelines
2. A role named `kubeflow-pipelines-public` with access to the above configMap
3. A rolebinding named `kubeflow-pipelines-public` which grants this role to all authenticated users

As of the writing of this comment, this value of `kubeflow_pipelines_version` is used neither in `pipelines` or in `manifests`, and that's why I'm not including it in the changes.

It seems that there is a plan for ensuring version compatibility when using the `kfp` sdk, see [this PR](https://github.com/kubeflow/sdk/issues/221).

NohaIhab: Thanks for clarifying @mvlassis, since it has no functionality yet we can skip it for this version, but let's keep an eye on it for future releases

charms/kfp-api/src/charm.py
NohaIhab: +1 from my side, @misohu to also review

NohaIhab (approved):

misohu (approved):

Issue #866, opened 2026-02-18
Allow `kfp` deployment on any namespace


dariofaccin:
### Context

As part of https://github.com/canonical/bundle-kubeflow/issues/1385, we want to allow the deployment of charmed Kubeflow on any namespace.
In this charm, the namespace `kubeflow` is hardcoded in multiple places.

## Profile controller configuration
It is possible to override the default `minio` provider configuration by defining a default endpoint like this:
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: kfp-launcher
  namespace: <a-namespace>
data:
  defaultPipelineRoot: <any>
  providers:
    minio:
      default:
        endpoint: <minio-host>:<minio-port>.<minio-namespace>
        credentials: 
          fromEnv: false
            secretRef:
              secretName: <minio_secret_name>
              accessKeyKey: MINIO_ACCESS_KEY
              secretKeyKey: MINIO_SECRET_KEY
```
The code of [`sync.py`](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-profile-controller/files/upstream/sync.py) has to be updated so that:

1. a new variable `minio_secret_name` is passed as parameter: this is already available at charm [source](https://github.com/canonical/kfp-operators/blob/1c5c1fccfd25e9f421c2f031bb8b48a0e7bbab64/charms/kfp-profile-controller/src/charm.py#L174)
2. `desired_resources` list always contain the `kfp-launcher` `ConfigMap`
3. such `ConfigMap` is populated, under the `data.providers`, with the information on minio service
4. the `defaultPipelineRoot` key updates the `data` map without deleting any other information

## Model name validation
In all charms, a method named `_check_model_name()`  or a `ModelNameGateComponent` are defined in order to prevent the charm to be deployed in a namespace different from `kubeflow`.
These clauses have to be removed.

## Authorizations
During the initial exploration, we faced some issues with RBAC and roles. In particular, API clients get this error when trying to list any resource:
```
[...]

PR #901, opened 2026-05-12
feat(kfp-profile-controller): allow deployment on any namespace

### LINKED ISSUE #389: Upstream KFP has hardcoded pointer to `minio-service.kubeflow` for object storage
See #387 for more background.

dariofaccin:
This PR adds the capability to deploy KFP profile controller on any namespace.

## Metadata gRPC service host
This value is now dynamically created from the charm model.
Unit test is updated accordingly.

## KFP API principal
The configuration option `kfp-api-principal` is removed and its value is now dynamically created from the charm model.
Related unit test is removed.

Ref issue: #866.

This PR closes #389.

### DIFF
--- a/charms/kfp-profile-controller/config.yaml
+++ b/charms/kfp-profile-controller/config.yaml
@@ -22,9 +22,4 @@ options:
 
       Note that manual deletion of old ConfigMaps is required after updating this, refer to:
       https://github.com/kubeflow/manifests/blob/v1.11.0/applications/pipeline/upstream/base/installs/generic/pipeline-install-config.yaml#L40-L42
-  kfp-api-principal:
-    type: string
-    default: "cluster.local/ns/kubeflow/sa/kfp-api"
-    description: >
-      The principal for the KFP API service account.
-      This is used in the AuthorizationPolicy that gets created in the sync.py script.
+
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -217,7 +217,7 @@ class KfpProfileControllerOperator(CharmBase):
                     VISUALIZATION_SERVER_TAG=self.images["visualization_server__version"],
                     FRONTEND_IMAGE=self.images["frontend__image"],
                     FRONTEND_TAG=self.images["frontend__version"],
-                    KFP_API_PRINCIPAL=str(self.model.config["kfp-api-principal"]),
+                    KFP_API_PRINCIPAL=f"cluster.local/ns/{self.model.name}/sa/kfp-api",
                     AMBIENT_ENABLED=self.service_mesh_component.component.ambient_mesh_enabled,
                     HOOKS_PATH=HOOKS_PATH,
                 ),
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -59,7 +59,6 @@ KFP_IMAGES_VERSION = "2.16.0"  # Remember to change this version also in default
 # This service name must be the Service from the mlmd-operator
 # FIXME: leaving it hardcoded now, but we should share this
 # host and port through relation data
-METADATA_GRPC_SERVICE_HOST = "metadata-grpc-service.kubeflow"
 METADATA_GRPC_SERVICE_PORT = "8080"
 NAMESPACE_LABEL = "pipelines.kubeflow.org/enabled"
 SYNC_CODE_FILE = Path("files/upstream/sync.py")
@@ -116,6 +115,7 @@ class KfpProfileControllerOperator(CharmBase):
         self.default_pipeline_root = self.model.config["default_pipeline_root"]
 
         self._container_name = next(iter(self.meta.containers))
+        self.metadata_grpc_service_host = f"metadata-grpc-service.{self.model.name}"
 
         # expose controller's port
         http_port = ServicePort(K8S_SVC_CONTROLLER_PORT, name="http", targetPort=CONTROLLER_PORT)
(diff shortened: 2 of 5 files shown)


charms/kfp-profile-controller/config.yaml
```diff
@@ -22,9 +22,4 @@ options:
 
       Note that manual deletion of old ConfigMaps is required after updating this, refer to:
       https://github.com/kubeflow/manifests/blob/v1.11.0/applications/pipeline/upstream/base/installs/generic/pipeline-install-config.yaml#L40-L42
-  kfp-api-principal:
```
MattiaSarti, line 25: note: this makes the pull request backward-incompatible, as we no longer have the option of changing the principal completely

charms/kfp-profile-controller/src/charm.py
```diff
@@ -217,7 +217,7 @@ def __init__(self, *args):
                     VISUALIZATION_SERVER_TAG=self.images["visualization_server__version"],
                     FRONTEND_IMAGE=self.images["frontend__image"],
                     FRONTEND_TAG=self.images["frontend__version"],
-                    KFP_API_PRINCIPAL=str(self.model.config["kfp-api-principal"]),
+                    KFP_API_PRINCIPAL=f"cluster.local/ns/{self.model.name}/sa/kfp-api",
```
MattiaSarti, line 220: thought: this hardcodes the trust domain (`cluster.local`) and the service account name (`kfp-api`) of the SPIFFE identity for the AuthorizationPolicy, which does not seem unreasonable, but do you know, @NohaIhab, if this could cause problems for general use cases? The service account name should already be dictated by the charm, right? And the trust domain is always expected to be the local cluster for internal usage only, correct? I don't know if we had broader use cases in mind when introducing ambient in [this pull request](https://github.com/canonical/kfp-operators/pull/820), where the whole principal was parametrized instead of its namespace only

charms/kfp-profile-controller/config.yaml
dariofaccin: This was discussed with @NohaIhab and since we never used or exposed this configuration option yet, it seemed reasonable to remove it.

NohaIhab: Yes, for the `kfp-api-principal` we are letting fall back to the default in [the ambient solution](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.11/modules/kubeflow-ambient/applications.tf), and the change here is behaving similar to the default when it comes to kubeflow namespace, while adding the option for other namespaces

charms/kfp-profile-controller/src/charm.py
mvlassis: Regarding the domain name, we will probably add this as a config option to follow the [upstream variable](https://github.com/kubeflow/pipelines/blob/504d7797c1e061327ddb2bcb8bc1c5dd4940cfc3/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml#L130-L131) in https://github.com/canonical/kfp-operators/issues/822

NohaIhab: I agree with you on the trust domain, but the service account name could change if someone decides to deploy `kfp-api` with a different name like
```
juju deploy kfp-api my-kfp-api ...
```
However, this is an issue that has been there and isn't introduced by this PR. Ideally, the SA name would be passed through a relation, but I believe this is beyond the scope of the namespace changes.
@dariofaccin Can you check if there's an issue open on this one, and if not file one?

mvlassis (approved):

charms/kfp-profile-controller/config.yaml
MattiaSarti: So this mean that we don't have to worry about the fact it is backward-incompatible because never exposed, right?

charms/kfp-profile-controller/src/charm.py
dariofaccin: @NohaIhab I opened [this issue](https://github.com/canonical/kfp-operators/issues/904) to track the Service Account issue. Regarding the cluster domain, I agree we can include it in the #822 effort.

MattiaSarti: @mvlassis, are we sure this:

> Regarding the domain name, we will probably add this as a config option to follow the [upstream variable](https://github.com/kubeflow/pipelines/blob/504d7797c1e061327ddb2bcb8bc1c5dd4940cfc3/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml#L130-L131) in https://github.com/canonical/kfp-operators/issues/822

is used also for principals in `AuthorizationPolicy`s? Just to double-check

charms/kfp-profile-controller/config.yaml
NohaIhab: yes, exactly @MattiaSarti

charms/kfp-profile-controller/src/charm.py
mvlassis: The question here is what `AuthorizationPolicy` we are referring to. The `KFP_API_PRINCIPAL` environment variable doesn't exist in upstream, and this makes sense since it is [CKF only](https://github.com/canonical/kfp-operators/blob/2a8f1aa70b49b53baccd60ee0ddec6728f4e7d58/charms/kfp-profile-controller/files/upstream/sync.py#L142-L144), right?

The [PR that introduced the change](https://github.com/kubeflow/pipelines/pull/12701) doesn't remove the hardcoding for the `AuthorizationPolicies` defined [here](https://github.com/kubeflow/pipelines/blob/289267879edc94c7a61ce91cb875ff2acbe28329/manifests/kustomize/base/installs/multi-user/istio-authorization-config.yaml#L39-L45), and the same is reflected in [kubeflow/manifests](https://github.com/kubeflow/manifests/blob/c1a557b3b3998ac5f4cb4947b1935bd80fd1364e/applications/pipeline/upstream/base/installs/multi-user/istio-authorization-config.yaml#L40-L45). I'm not sure whether this is an omission or not. It seems to me that since the cluster domain is configurable in the `ml-pipeline` deployment [here](https://github.com/kubeflow/pipelines/blob/289267879edc94c7a61ce91cb875ff2acbe28329/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml#L130), shouldn't the change be reflected in the `AuthorizationPolicy` as well?

MattiaSarti: Thanks for the details, @mvlassis.

Actually, this is what I was trying to understand in the first place:

> shouldn't the change be reflected in the AuthorizationPolicy as well?

But since you showed all those principals are kind of hardcoded, as you referenced, then we should be safe to assume our principals won't deviate either, and we can therefore hardcode the trust domain as @dariofaccin was doing.

MattiaSarti (approved):

PR #947, opened 2026-06-16
fix: reintroduce kfp-api-principal config

### LINKED ISSUE #946: Re-introduce config option for principal for compatibility
### Bug Description
In https://github.com/canonical/kfp-operators/pull/914, we introduced a commit for `track/2.16` that removes the `kfp-api-principal` from the config of `kfp-profile-controller`. 

### Environment
- `track/2.16` of `kfp-profile-controller`, although this change shouldn't be present in `2.16/edge` due to the release action failing

### Additional Context
We should:
- Keep both config options
- Add a deprecation message
- Backport to `track/2.16`

dariofaccin:
This PR reintroduces `kfp-api-principal` config option, allowing the refresh from older versions of the `kfp-profile-controller` charm where this option might be set.

If the configuration option is set, then its value is used to populate the env variable `KFP_API_PRINCIPAL`.
If the configuration option is not set, then the env variable `KFP_API_PRINCIPAL` is dinamically constructed by the charm using the model name and the `kfp_api_service_account_name` config option.

**Note**: setting a custom `kfp_api_service_account_name` and `kfp-api-principal` results in the charm going into Blocked status.

This PR closes #946.

### DIFF
--- a/charms/kfp-profile-controller/src/charm.py
+++ b/charms/kfp-profile-controller/src/charm.py
@@ -62,8 +62,6 @@ KFP_IMAGES_VERSION = "2.16.0"  # Remember to change this version also in default
 METADATA_GRPC_SERVICE_PORT = "8080"
 NAMESPACE_LABEL = "pipelines.kubeflow.org/enabled"
 SYNC_CODE_FILE = Path("files/upstream/sync.py")
-# Default value of the `kfp_api_service_account_name` config option, must match config.yaml
-DEFAULT_KFP_API_SERVICE_ACCOUNT_NAME = "kfp-api"
 
 HOOKS_PATH = Path("/var/lib/pebble/default")
 
@@ -112,7 +110,7 @@ class KfpProfileControllerOperator(CharmBase):
                 parse_images_config(self.model.config["custom_images"]),
             )
             # Validate the principal-related config early so the unit goes to BlockedStatus
-            # if both `kfp-api-principal` and `kfp_api_service_account_name` are set.
+            # if both `kfp-api-principal` and `kfp_api_service_account_name` are non-empty.
             self._get_kfp_api_principal()
         except ErrorWithStatus as e:
             self.unit.status = e.status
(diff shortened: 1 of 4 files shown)


charms/kfp-profile-controller/src/charm.py
```diff
@@ -109,6 +111,9 @@ def __init__(self, *args):
                 DEFAULT_IMAGES,
                 parse_images_config(self.model.config["custom_images"]),
             )
+            # Validate the principal-related config early so the unit goes to BlockedStatus
+            # if both `kfp-api-principal` and `kfp_api_service_account_name` are set.
+            self._get_kfp_api_principal()
```
MattiaSarti, line 114: note: I guess it is fine to keep using this mixed pattern of both Chisme-fashion components and direct status handling (as already done for image configs) for these changes, since they are temporary

charms/kfp-profile-controller/src/charm.py
```diff
@@ -234,6 +238,36 @@ def __init__(self, *args):
...
+        kfp_api_service_account_name = self.model.config["kfp_api_service_account_name"]
+
+        if kfp_api_principal:
+            if kfp_api_service_account_name != DEFAULT_KFP_API_SERVICE_ACCOUNT_NAME:
+                raise ErrorWithStatus(
+                    "Cannot set both `kfp-api-principal` and `kfp_api_service_account_name`. "
+                    "Use only `kfp_api_service_account_name` as `kfp-api-principal` is "
+                    "deprecated.",
+                    BlockedStatus,
+                )
```
MattiaSarti: issue: this might be confusing to the end user - how about explicitly stating that we want them to set the new config option back to the default value? Or, even better, how about having them set the new config option to an empty string, when the old one is set? So that we avoid them having two different, non-empty config options at the same time

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -327,6 +327,61 @@ def test_pebble_services_updated_when_kfp_api_service_account_name_changes(
...
+def test_pebble_services_with_deprecated_kfp_api_principal(
+    harness: Harness,
+    mocked_lightkube_client,
+    mocked_kubernetes_service_patch,
+):
+    """Test that the deprecated `kfp-api-principal` config is used directly when set."""
+    # Arrange
+    custom_principal = "cluster.local/ns/some-namespace/sa/some-sa"
+    harness.update_config({"kfp-api-principal": custom_principal})
+    expected_environment = generate_expected_environment(harness.model.name)
```
MattiaSarti, line 342: issue: doesn't the expected environment have the default value for `kfp_api_sa` in the principal, that is `kfp-api`, which is different from the `some-sa` used in this test case?

charms/kfp-profile-controller/tests/integration/test_charm_ambient.py
```diff
@@ -204,6 +204,25 @@ async def test_insecure_authorization_policy_is_missing(
     assert excinfo.value.response.status_code == 404
 
 
+# Targeted for ambient integration
```
MattiaSarti, line 207: praise: nice test idea

MattiaSarti (changes requested):

charms/kfp-profile-controller/tests/unit/test_operator.py
dariofaccin: We are overriding the generated one in L340: `expected_environment["KFP_API_PRINCIPAL"] = custom_principal`

MattiaSarti (approved):

charms/kfp-profile-controller/tests/unit/test_operator.py
MattiaSarti: Right, my apologies!

misohu (approved):
Really nice quick fix

PR #955, opened 2026-06-18
feat: support s3-integrator as an object storage backend

mvlassis:
## Summary
- Add `s3` interface support (via an `s3-credentials` relation) to **kfp-api**, **kfp-ui** and **kfp-profile-controller**, letting them use an [s3-integrator](https://charmhub.io/s3-integrator) as an alternative to the in-cluster MinIO (`object-storage`) backend.
- Exactly one of `object-storage` or `s3-credentials` may be related at a time; relating both blocks the charm (enforced by a `RelationCountGateComponent` conflict gate in kfp-ui/kfp-profile-controller, and inline in kfp-api).
- The s3 endpoint is parsed into the host, port and TLS flag the workloads expect. The kfp-profile-controller `sync.py` hook now derives SSL/region and omits the `.namespace` suffix when using s3.

## Changes
- **charm.py** (×3): normalized `_get_object_storage_data()` + `_parse_s3_endpoint()` supporting both backends. kfp-api uses `object_storage.S3Requirer`; kfp-ui/kfp-profile-controller use chisme `S3RequirerComponent` + `RelationCountGateComponent`.
- **Workload**: profile-controller `sync.py` (`minio_ssl`/`minio_region`, empty-namespace endpoint) and `pebble_components.py` (`MINIO_SSL`/`MINIO_REGION` env, dropped namespace-required check).
- **metadata.yaml** (×3): new `s3-credentials` relation (interface `s3`).
- **Deps**: chisme `^0.4.27` + `object-storage-charmlib ^1.0.0` (pyproject + poetry.lock).
- **Tests**: integration `test_charm.py` → `test_charm_object_storage.py` (pure rename) + new `test_charm_s3.py` (×3); new unit tests for endpoint parsing, conflict detection, waiting status, and s3-backed pebble services.
- **CI/tox**: new `integration-s3` tox env (root + per-charm), S3 passenv, and `integration-s3` CI matrix entry with excludes for charms that don't integrate with s3.

## Behavior change
- kfp-api's `OBJECTSTORECONFIG_SECURE` now derives from the `object-storage` relation's `secure` field instead of being hardcoded to `false`.

Closes #902

PR #956, opened 2026-06-18
feat(kfp-api): add s3-credentials relation to relate with s3-integrator

### LINKED ISSUE #964: Add `s3` interface to `kfp-api` to relate with `s3-integrator`
Part of #902.

### Context

.

### What needs to get done

.

### Definition of Done

.

### LINKED ISSUE #9689: [feature] remove hard-coded configs in KFP V2 launcher
### Feature Area

/area backend

## Problem

Currently, we are hardcoding a few important configs in KFP V2's launcher and cache code.
This is making it difficult for distributions to support KFP v2 in a generic way.

__Launcher - The minio credentials secret is hard-coded as `mlpipeline-minio-artifact`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/objectstore/object_store.go#L292

__Launcher - The minio credentials secret keys are hard-coded as `accesskey` and `secretkey`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/objectstore/object_store.go#L324-L325

__Launcher - The minio endpoint is hard-coded as `minio-service.kubeflow:9000`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/objectstore/object_store.go#L291

__Cache - The `ml-pipeline` endpoint is hard-coded as `ml-pipeline.kubeflow:8887`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/cacheutils/cache.go#L26

## Solution

We want to allow the above configs to be changed from their defaults, preferably on a per-namespace basis.

I propose we add these configs to the `ConfigMap/kfp-launcher` which already exists in each profile namespace to set `defaultPipelineRoot`.

 For example, a new `ConfigMap/kfp-launcher` in a profile namespace might look like this:

```yaml
data:
  defaultPipelineRoot: "minio://mlpipeline/v2/artifacts"
 
  ## minio endpoint config
  minioEndpoint: "minio.example.com:9000"
 
  ## minio auth configs
  minioAuthSecret: "mlpipeline-minio-artifact"
  minioAuthSecretAccessKeyKey: "access_key"
  minioAuthSecretAccessKeyKey: "access_key"

  ## ml pipeline endpoint config
  mlPipelineEndpoint: "ml-pipeline.kubeflow:8887"
```

The question now becomes how to propagate these configs to the `kfp-launcher` container in each workflow Pod.
I think there are two options:

- OPTION 1: have the `kfp-launcher` container read the `ConfigMap/kfp-launcher`
    - [when it executes](https://github.com/kubeflow/pipelines/blob/dbebbde2001f7d2a7372a8abfd9c5c5a0eaea2b8/backend/src/v2/component/launcher_v2.go#L132), it can pass the minio configs down as it calls [`objectstore.OpenBucket`](https://github.com/kubeflow/pipelines/blob/dbebbde2001f7d2a7372a8abfd9c5c5a0eaea2b8/backend/src/v2/component/launcher_v2.go#L163) 
    - setting `mlPipelineEndpoint` has to be done before it runs [`cacheutils.NewClient()`](https://github.com/kubeflow/pipelines/blob/dbebbde2001f7d2a7372a8abfd9c5c5a0eaea2b8/backend/src/v2/component/launcher_v2.go#L115)
- OPTION 2:  read the `ConfigMap/kfp-launcher` in [`RootDAG` at the same time as we read the `defaultPipelineRoot`](https://github.com/kubeflow/pipelines/blob/eadeb3096b9cb3d3c8495491a461154b973b4e2a/backend/src/v2/driver/driver.go#L146)
    - (this approach reduces the number of API calls to read `ConfigMap/kfp-launcher`)

## Other Thoughts

We could make this slightly more advanced to allow for "selecting" credentials based on the pipelines, because not all pipelines have to use the default [bucket_root](https://kubeflow-pipelines.readthedocs.io/en/latest/source/client.html?h=root#kfp.client.Client.run_pipeline.pipeline_root), so might need different credentials.

For example, we could extend `ConfigMap/kfp-launcher` like this:

```yaml
data:
  defaultPipelineRoot: "minio://my-bucket/v2/artifacts"

  ## minio endpoint config
  minioEndpoint: "minio.example.com:9000"

  ## ml pipeline endpoint config
  mlPipelineEndpoint: "ml-pipeline.kubeflow:8887"

  ## bucket auth configs (as a YAML string)
  bucketAuth: |
    ## list of "auth providers" for minio
    minio:

      ## specifies auth for objects under "minio://my-bucket/v2/artifacts/"
      - bucketName: "my-bucket"
        keyPrefix: "v2/artifacts/"
        authFromSecret:
          secretName: "my-secret-in-profile-namespace"
          accessKeyKey: "access_key"
          secretKeyKey: "secret_key"

    ## similar, but for s3 buckets
    s3: []

    ## similar, but for gcs buckets
    gcs: []

```

---

Love this idea? Give it a 👍.

mvlassis:
## Summary
- Add `s3-credentials` relation using the `s3` interface.
- Implemented the `RelationCountGate` component manually
- The s3 endpoint is parsed into the host, port and TLS flag kfp expects (`OBJECTSTORECONFIG_*`)
- Update the S3 service to include the `tls-ca-chain`, similar to the structure for the S3 manager in [object-storage-integrators](https://github.com/canonical/object-storage-integrator/blob/main/s3/src/managers/s3.py)
- Add integration tests, split into 2 distinct files:
  - `test_charm_s3.py`, which tests the relation with `s3-integrator:s3-credentials`
  - `test_charm_object_storage.py`, which tests the relation with `minio:object-storage`
- Add test for migrating from `minio-operator` to `s3-integrator` in `integration-object-storage`

## Removal
This PR removes the `minio-service` service created by [kfp-api](https://github.com/canonical/kfp-operators/blob/f0df861a2bd49819dd9abf7097994f8b91898de4/charms/kfp-api/src/templates/minio-service.yaml.j2). The context is:
- This PR was initially created to fix https://github.com/kubeflow/pipelines/issues/9689, which has been fixed
- The fix is the introduction of a `kfp-launcher` configMap that can specify the endpoint (among others) instead of using the default hardcoded value
- As of the merge of https://github.com/canonical/kfp-operators/pull/870, we now always set the endpoint in `kfp-launcher` to override the default value, which has been `http://seaweedfs.kubeflow:8333` as of Kubeflow pipelines [2.15.0](https://github.com/kubeflow/pipelines/blob/8fe8d7bdd88f027bde080f619762a0b3294796db/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py)

## Notes
- Split out of #955. Companion PRs add the same support to kfp-ui and kfp-profile-controller.
- The root `tox.ini` change is identical across the three split PRs (conflict-free); the `ci.yaml` matrix enables `integration-s3` only for kfp-api here, so merging the companion PRs will require a small exclude-list merge.
- I removed all references of `mariadb` and replaced them with `mysql-k8s` :)

Closes #964


tox.ini
```diff
@@ -3,7 +3,7 @@
 [tox]
 skipsdist = True
 skip_missing_interpreters = True
-envlist = {kfp-api,kfp-metadata-writer,kfp-persistence,kfp-profile-controller,kfp-schedwf,kfp-ui,kfp-viewer,kfp-viz}-{lint,unit,integration,integration-ambient},bundle-integration
+envlist = {kfp-api,kfp-metadata-writer,kfp-persistence,kfp-profile-controller,kfp-schedwf,kfp-ui,kfp-viewer,kfp-viz}-{lint,unit,integration,integration-ambient,integration-object-storage},bundle-integration
```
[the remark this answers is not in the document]
mvlassis: Note that we'll also update this in the other PRs

charms/kfp-api/src/charm.py
```diff
@@ -551,6 +566,84 @@ def _get_object_storage(self, interfaces):
...
+                "access-key": data["access-key"],
+                "secret-key": data["secret-key"],
+                "host": host,
+                "port": port,
+                "secure": secure,
+                "region": data.get("region", ""),
+                "bucket": data.get("bucket", ""),
+                "tls-ca-chain": data.get("tls-ca-chain"),
+                "is_s3": True,
+            }
```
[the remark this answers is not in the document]
mvlassis: - `minio` always returns a Kubernetes DNS name
- At least one relation out of the 2 IS required, we just don't want to block on either on them, and use the `RelationGateCountComponent` to set the proper application status

Dismissing comments

charms/kfp-api/src/services/s3.py
```diff
@@ -15,23 +17,49 @@ class S3BucketWrapper:
...
+            tmp.close()
+            self._ca_file = tmp.name
 
         self._client: botocore.client.BaseClient = None
 
+    def __del__(self):
+        """Clean up the temporary CA file if one was created."""
+        if self._ca_file and os.path.exists(self._ca_file):
+            os.unlink(self._ca_file)
+
```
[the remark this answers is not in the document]
mvlassis: Already addressed

charms/kfp-api/src/charm.py
```diff
@@ -551,6 +566,88 @@ def _get_object_storage(self, interfaces):
...
+
+        Raises:
+            ErrorWithStatus(..., Waiting) if the relation exists but required data
+                (access-key, secret-key, endpoint) is not yet available.
+        """
+        relation = self.model.get_relation("s3-credentials")
+        info = self.s3.get_storage_connection_info(relation)
+        required_fields = ("access-key", "secret-key", "endpoint")
+        if not info or not all(info.get(field) for field in required_fields):
+            raise ErrorWithStatus("Waiting for s3-credentials relation data", WaitingStatus)
```
NohaIhab: suggestion(non-blocking): make the message tailored to whether all info is missing or a specific field, if any. It can help with debugging.

charms/kfp-api/src/charm.py
```diff
@@ -551,6 +566,88 @@ def _get_object_storage(self, interfaces):
...
+
+    @staticmethod
+    def _parse_s3_endpoint(endpoint: str) -> tuple:
+        """Parse an s3 endpoint into a (host, port, secure) tuple.
+
+        The endpoint may be a full URL (e.g. "https://s3.example.com:443") or a bare
+        "host[:port]". kfp expects the host, port and TLS flag as separate values, so the
+        URL scheme (when present) determines the default port and whether TLS is used.
+        """
+        parsed = urlparse(endpoint if "://" in endpoint else f"//{endpoint}")
```
NohaIhab: suggestion: rename to `parsed_endpoint` or something more clear

charms/kfp-api/src/charm.py
```diff
@@ -894,6 +993,10 @@ def _ensure_bucket_exists(self) -> None:
             s3_wrapper.create_bucket(bucket_name)
             return
 
+        except botocore.exceptions.SSLError as e:
+            msg = "Object storage TLS verification failed. Check CA chain configuration."
+            self.logger.warning(f"{msg}: {e}")
+            raise ErrorWithStatus(msg, WaitingStatus)
```
NohaIhab: question: what do we expect to resolve the status here? Just wondering if `WaitingStatus` is appropriate

charms/kfp-api/tests/integration/test_charm_object_storage.py
NohaIhab, line 1: suggestion(non-blocking): I realize it's copied from the other test file, but we should not introduce new tests with mariadb.

NohaIhab (commented):
thanks @mvlassis, left some comments

charms/kfp-api/src/charm.py
mvlassis: Updated https://github.com/canonical/kfp-operators/pull/956/commits/7f33cbeddfd53a0f4ff5022b9fced1b08b116cf6 to
- Display a specific message when everything is missing
- Otherwise, display the missing *required* fields

charms/kfp-api/src/charm.py
mvlassis: https://github.com/canonical/kfp-operators/pull/956/commits/11b7726e13332fd7c2e405f341d44dbdb6a7e275

charms/kfp-api/src/charm.py
mvlassis: I wasn't sure about this either, I guess a TLS verification error can't be resolved on it's own... I'll change this to a blocked status

mvlassis: https://github.com/canonical/kfp-operators/pull/956/commits/684dcd0dff7185d382673cd512868ac4d0c44e14

charms/kfp-api/tests/integration/test_charm_object_storage.py
mvlassis: It's way overdue, I replaced all instances-references of `mariadb` with `mysql` in https://github.com/canonical/kfp-operators/pull/956/commits/5698acb8fde3ff9a99edc15e720ff5277a15df25. That makes sense since all deployments we do are with `mysql-k8s`, so we should test proper behavior with that charm

charms/kfp-api/tox.ini
```diff
@@ -80,8 +85,15 @@ commands_pre =
...
-description = Run integration tests
+commands = pytest -vv --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_charm_s3.py --log-cli-level=INFO -s {posargs}
+description = Run integration tests with s3-integrator via the s3-credentials relation
+commands_pre = 
+	poetry install --only integration
+skip_install = true
+
+[testenv:integration-object-storage]
+commands = pytest -vv --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_charm_object_storage.py --log-cli-level=INFO -s {posargs}
+description = Run integration tests with minio-operator via the object-storage relation
```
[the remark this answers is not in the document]
mvlassis: Not necessary, dismissing

charms/kfp-api/src/services/s3.py
```diff
@@ -15,23 +17,56 @@ class S3BucketWrapper:
...
+        self.region: str = region
+
+        self._ca_file: Optional[str] = None
+        if tls_ca_chain:
+            ca_chain_pem = "\n".join(tls_ca_chain)
+            tmp = tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".pem")
+            tmp.write(ca_chain_pem)
+            tmp.flush()
+            tmp.close()
+            self._ca_file = tmp.name
```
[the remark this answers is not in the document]
mvlassis: This is not relevant for this case, since charm hooks are supposed to be short-running processes

charms/kfp-api/tests/integration/test_charm_object_storage.py
NohaIhab: awesome, thank you!

NohaIhab (approved):
thanks @mvlassis

charms/kfp-api/src/services/s3.py
```diff
@@ -15,23 +17,56 @@ class S3BucketWrapper:
...
+        self.secure: bool = secure
+        self.region: str = region
+
+        self._ca_file: Optional[str] = None
+        if tls_ca_chain:
+            ca_chain_pem = "\n".join(tls_ca_chain)
+            tmp = tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".pem")
+            tmp.write(ca_chain_pem)
+            tmp.flush()
+            tmp.close()
```
dariofaccin:
```suggestion
            with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".pem") as fd:
            fd.write(ca_chain_pem)
```
suggestion(non-blocking): why not use a context manager? We won't need explicit `flush()` and `close()` (which in python 3 [invokes `flush()` anyway](https://docs.python.org/3/library/io.html#io.IOBase.close)).

charms/kfp-api/src/charm.py
```diff
@@ -894,6 +1000,10 @@ def _ensure_bucket_exists(self) -> None:
             s3_wrapper.create_bucket(bucket_name)
             return
 
+        except botocore.exceptions.SSLError as e:
+            msg = "Object storage TLS verification failed. Check CA chain configuration."
+            self.logger.warning(f"{msg}: {e}")
```
dariofaccin: suggestion(non-blocking): I'd use `error` level given it blocks the charm.

charms/kfp-api/src/charm.py
```diff
@@ -551,6 +566,95 @@ def _get_object_storage(self, interfaces):
...
+
+        Exactly one of the `object-storage` or `s3-credentials` relations is expected.
+
+        Raises:
+            ErrorWithStatus(..., Blocked) if both relations are established at once.
+            ErrorWithStatus(..., Blocked) if neither relation is established (raised by the
+                object-storage SDI validation).
+            ErrorWithStatus(..., Waiting) if the active relation has no data yet.
+        """
+        if self.model.relations["object-storage"] and self.model.relations["s3-credentials"]:
```
dariofaccin: question: what happens if we don't relate neither with `object-storage` or `s3-credentials`?

charms/kfp-api/src/charm.py
dariofaccin, line 505: issue(blocking): this relation is not mandatory anymore.

charms/kfp-api/src/charm.py
```diff
@@ -871,18 +975,20 @@ def _check_config(self) -> None:
...
+            s3_service=obj["host"],
             s3_port=obj["port"],
+            secure=obj["secure"],
+            region=obj["region"],
+            tls_ca_chain=obj.get("tls-ca-chain"),
         )
 
         # Try creating the bucket we need for object storage
-        bucket_name = self.model.config["object-store-bucket-name"]
+        bucket_name = obj["bucket"] or self.model.config["object-store-bucket-name"]
```
dariofaccin: suggestion(non-blocking): I'add a log (up to you the level) to inform the operator that the bucket name is empty and fallbacks to the config option.

charms/kfp-api/src/services/s3.py
```diff
@@ -15,23 +17,56 @@ class S3BucketWrapper:
...
+        try:
+            os.unlink(ca_file)
+        except FileNotFoundError:
+            pass
+        except OSError:
+            pass
+
     def create_bucket(self, bucket_name):
         """Create a bucket via the client with configured timeouts."""
         client = boto3.client(
```
dariofaccin: question: why are we instantiating a new client instead of using the cached one with `self.client`?

charms/kfp-api/src/services/s3.py
dariofaccin: question: why are we instantiating a new client instead of using the cached one with `self.client`?

charms/kfp-api/src/charm.py
```diff
@@ -551,6 +566,95 @@ def _get_object_storage(self, interfaces):
...
+    def _parse_s3_endpoint(endpoint: str) -> tuple:
+        """Parse an s3 endpoint into a (host, port, secure) tuple.
+
+        The endpoint may be a full URL (e.g. "https://s3.example.com:443") or a bare
+        "host[:port]". kfp expects the host, port and TLS flag as separate values, so the
+        URL scheme (when present) determines the default port and whether TLS is used.
+        """
+        parsed_endpoint = urlparse(endpoint if "://" in endpoint else f"//{endpoint}")
+        secure = parsed_endpoint.scheme == "https"
+        port = parsed_endpoint.port or (443 if secure else 80)
```
dariofaccin: issue(blocking): splitting the hair here, but for the following endpoint `s3.example.com:443` I'd expect the `secure` variable to be `True`

charms/kfp-api/tests/unit/test_operator.py
```diff
@@ -1140,3 +1140,210 @@ def test_object_storage_timeout(
...
+        "endpoint,expected",
+        [
+            # full URLs: scheme decides the TLS flag and default port
+            ("http://10.0.0.1", ("10.0.0.1", 80, False)),
+            ("https://s3.example.com", ("s3.example.com", 443, True)),
+            ("http://10.0.0.1:9000", ("10.0.0.1", 9000, False)),
+            ("https://s3.example.com:443", ("s3.example.com", 443, True)),
+            # bare host[:port]: defaults to non-TLS
+            ("10.0.0.1", ("10.0.0.1", 80, False)),
+            ("minio.example:9000", ("minio.example", 9000, False)),
```
dariofaccin, line 1126: suggestion: add a test case for `s3.amazon.com:443`.

dariofaccin (changes requested):
Left some suggestions and questions.
In the integration tests we are deploying `grafana-agent`, checking alert rules (`test_alert_rules`), metrics (`test_metrics_endpoint`), logging (`test_logging`) and container security context (`test_container_security_context`) in both `object-storage` and `s3` scenarios. Does it provide more coverage or is it CI time wasted?

charms/kfp-api/src/charm.py
mvlassis: Handled in https://github.com/canonical/kfp-operators/pull/956/commits/0058b73d881eb8e73f2fad645b900848416fba7d

charms/kfp-api/tests/unit/test_operator.py
mvlassis: Handled in https://github.com/canonical/kfp-operators/pull/956/commits/0058b73d881eb8e73f2fad645b900848416fba7d

charms/kfp-api/src/charm.py
mvlassis: Addressed

mvlassis:
We discussed this question by @dariofaccin:
> In the integration tests we are deploying grafana-agent, checking alert rules (test_alert_rules), metrics (test_metrics_endpoint), logging (test_logging) and container security context (test_container_security_context) in both object-storage and s3 scenarios. Does it provide more coverage or is it CI time wasted?

And the decision was to keep these tests, as it is the same pattern we use for the `test_charm_ambient.py` tests. The future plan is to deprecate the `object-storage` integration in the future, and thus remove the `test_charm_object_storage.py` file completely.

charms/kfp-api/src/services/s3.py
mvlassis: Updated in https://github.com/canonical/kfp-operators/pull/956/commits/1102e34da3215936d09d466667d2471454fec64f (we're using the singleton pattern already)

charms/kfp-api/src/services/s3.py
mvlassis: Updated in https://github.com/canonical/kfp-operators/pull/956/commits/1102e34da3215936d09d466667d2471454fec64f

charms/kfp-api/src/charm.py
mvlassis: This is an interesting point, we should actually never reach the *OR* here. The requirer is defined:
```
        self.s3 = S3Requirer(
            self,
            relation_name="s3-credentials",
            bucket=self.model.config["object-store-bucket-name"],
        )
```
This means that the `object-store-bucket-name` is passed to the `s3-integrator`, which returns:
- The bucket on the *provider* side if one is provided in `s3-integrator`
- The bucket from the *requirer* side (meaning this config option) if no `bucket` is specified on the provider side

So in both cases, a bucket is provided. A very edge case is if both the `requirer` and `provider` value have an empty bucket, which we should handle properly.

mvlassis: I added https://github.com/canonical/kfp-operators/pull/956/commits/d164ce55bbb53908f1be169abbed5c976ab51272 to:
- Log *info* if the config is used
- Emit error if there is no bucket specified on either side

charms/kfp-api/src/charm.py
mvlassis: Added in https://github.com/canonical/kfp-operators/pull/956/commits/d164ce55bbb53908f1be169abbed5c976ab51272, similar to how we will handle all charms

charms/kfp-api/src/charm.py
mvlassis: Addressed in https://github.com/canonical/kfp-operators/pull/956/commits/d164ce55bbb53908f1be169abbed5c976ab51272 (we want to block in this case, since the object storage *is* mandatory)

charms/kfp-api/src/charm.py
```diff
@@ -200,18 +211,21 @@ def container(self):
...
-        object_storage = self._get_object_storage(interfaces)
+        object_storage = self._get_object_storage_data()
 
-        minio_url = f"{object_storage['service']}.{object_storage['namespace']}.svc.cluster.local"
+        if object_storage["is_s3"]:
+            # The s3 endpoint is already an externally-resolvable host.
+            minio_url = object_storage["host"]
+        else:
+            # Must include .svc.cluster.local for in-cluster DNS resolution
+            minio_url = f"{object_storage['host']}.svc.cluster.local"
```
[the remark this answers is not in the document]
mvlassis: Should be handled by https://github.com/canonical/kfp-operators/pull/956/commits/26d21b379c5a7b2754b85cf40309c85260b75c36

dariofaccin (approved):
LGTM. Thanks.

NohaIhab (approved):

PR #957, opened 2026-06-18
feat(kfp-ui): add s3-credentials relation to relate with s3-integrator

mvlassis:
## Summary
- Add `s3` interface support (via an `s3-credentials` relation) to **kfp-profile-controller**, letting it use an [s3-integrator](https://charmhub.io/s3-integrator) as an alternative to the in-cluster MinIO (`object-storage`) backend.
- A `RelationCountGateComponent` conflict gate enforces that exactly one of `object-storage` or `s3-credentials` is related at a time (relating both, or neither, blocks the charm)).
- An `ObjectStorageValidator` component to pick the proper component depending on the current active object storage relation
- The s3 endpoint is parsed into the `MINIO_*` host, port and SSL flag the ml-pipeline-ui frontend expects, with an empty `MINIO_NAMESPACE` (s3 has no namespace concept).
- Now that all 3 components are finished, new bundle tests that use `s3-integrator` only (with the istio ambient charms)

## Notes
- Tried to harnomize according to the comments and changes introduced in https://github.com/canonical/kfp-operators/pull/958
- There is a quite ugly piece of code for using the `deploy_and_assert_s3_integrator` in the new bundle tests, we should update `charmed-kubeflow-chisme` to make this more `jubilant`-native in the future

## Notes
- Closes #965
- 1/3 of #955.

PR #958, opened 2026-06-18
feat(kfp-profile-controller): add s3-credentials relation to relate with s3-integrator

### LINKED ISSUE #966: Add `s3` interface to `kfp-profile-controller` to relate with `s3-integrator`
Part of #902.

### Context

.

### What needs to get done

.

### Definition of Done

.

mvlassis:
## Summary
- Add `s3` interface support (via an `s3-credentials` relation) to **kfp-profile-controller**, letting it use an [s3-integrator](https://charmhub.io/s3-integrator) as an alternative to the in-cluster MinIO (`object-storage`) backend.
- A `RelationCountGateComponent` conflict gate enforces that exactly one of `object-storage` or `s3-credentials` is related at a time (relating both, or neither, blocks the charm).
- Add three new `MINIO_*` arguments passed into the `sync.py` workload container

## Notes
- Split out of #955. 
- The root `tox.ini` change is identical across the three split PRs (conflict-free); the `ci.yaml` matrix enables `integration-s3` only for kfp-profile-controller here, so merging the companion PRs will require a small exclude-list merge.

Closes #966


charms/kfp-profile-controller/src/charm.py
```diff
@@ -197,45 +208,135 @@ def __init__(self, *args):
...
-                    HOOKS_PATH=HOOKS_PATH,
-                ),
+                inputs_getter=self._generate_kfp_profile_controller_inputs,
             ),
             depends_on=[
                 self.leadership_gate,
                 self.kubernetes_resources,
-                self.object_storage_relation,
+                self.s3_relations_conflict_detector,
                 self.service_mesh_component,
```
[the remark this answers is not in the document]
mvlassis: Resolved in https://github.com/canonical/kfp-operators/pull/958/commits/1cf22b1fa8598da0bf1c4fe9ac8aad2e2e04f3dc

charms/kfp-profile-controller/src/charm.py
```diff
@@ -197,45 +208,135 @@ def __init__(self, *args):
...
+                "region": data.get("region", ""),
+                # The s3 interface has no namespace concept, so the endpoint is just host:port.
+                "endpoint": f"{host}:{port}",
+            }
+        data = active.get_data()
+        # When minimum_related_applications != maximum_related_applications,
+        # SdiRelationDataReceiverComponent.get_data() returns a list of dicts rather than a
+        # single dict. Extract the first (and expected-only) entry.
+        if isinstance(data, list):
+            data = data[0]
```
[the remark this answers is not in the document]
mvlassis: Handled properly in https://github.com/canonical/kfp-operators/pull/958/commits/d844527b5fb1b5b8841e7e640673f9f82f7e92ef

charms/kfp-profile-controller/src/charm.py
```diff
@@ -197,45 +208,135 @@ def __init__(self, *args):
...
+        if isinstance(active, S3RequirerComponent):
+            # get_data() returns a list; exactly one S3 relation is expected (enforced by
+            # the conflict detector), so we take the first entry.
+            data = active.get_data()[0]
+            host, port, secure = self._parse_s3_endpoint(data["endpoint"])
+            if not host:
+                raise ErrorWithStatus(
+                    f"Invalid s3 endpoint: {data['endpoint']!r}",
+                    BlockedStatus,
+                )
```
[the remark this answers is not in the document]
mvlassis: Handled properly in https://github.com/canonical/kfp-operators/pull/958/commits/d844527b5fb1b5b8841e7e640673f9f82f7e92ef

charms/kfp-profile-controller/metadata.yaml
```diff
@@ -20,8 +20,14 @@ provides:
...
     interface: cross_model_mesh
 requires:
+  s3-credentials:
+    interface: s3
+    optional: true
+    limit: 1
   object-storage:
     interface: object-storage
+    optional: true
+    limit: 1
```
MattiaSarti, line 30: note: similar considerations apply as in https://github.com/canonical/kserve-operators/pull/488#discussion_r3475570193

.github/workflows/ci.yaml
```diff
@@ -207,11 +207,10 @@ jobs:
             test-type: integration-ambient
           - charm: kfp-metadata-writer
             test-type: integration-ambient
-          # TODO: Update with other charms that have an integration via object-storage
+          # integration-object-storage is only defined for the charms that integrate with
+          # s3-integrator, whose default `integration` env now runs the s3 backend test instead
```
NohaIhab: suggestion(non-blocking): Can we do the same as kserve and rename it to `integration-s3` to avoid confusion

charms/kfp-profile-controller/src/charm.py
```diff
@@ -236,6 +225,140 @@ def __init__(self, *args):
...
+        Supports both the `object-storage` and `s3` interfaces,
+        returning a common dict with keys: access_key, secret_key, host, namespace, port,
+        secure, region, endpoint.
+
+        Raises:
+            ErrorWithStatus(WaitingStatus): if the active relation exists but the
+                provider hasn't yet populated the required databag fields.
+            ErrorWithStatus(BlockedStatus): if the s3 endpoint is malformed.
+        """
+        active = self.active_storage_component
```
NohaIhab: nit: descriptive variable naming pls

charms/kfp-profile-controller/tests/integration/test_charm_s3.py
NohaIhab, line 1: suggestion(blocking): what about the kfp bundle [tests](https://github.com/canonical/kfp-operators/blob/main/tests/integration/test_kfp_functional.py)? will we update them to use s3, or it will be covered elsewhere? I think this is where you actually create the pipeline and pass the artefacts so it's important to test this with `s3-integrator`.

charms/kfp-profile-controller/src/charm.py
```diff
@@ -236,6 +225,140 @@ def __init__(self, *args):
...
+                    "Waiting for s3-credentials relation data, missing fields: "
+                    f"{', '.join(missing)}",
+                    WaitingStatus,
+                )
+            host, port, secure = self._parse_s3_endpoint(data["endpoint"])
+            if not host:
+                raise ErrorWithStatus(
+                    f"Invalid s3 endpoint: {data['endpoint']!r}",
+                    BlockedStatus,
+                )
```
NohaIhab: issue(blocking): It's not clear to me what's catching the `ErrorWithSatatus` and propagating the status. How does this actually work alongside the components? By raising errors for the status in the `charm.py` we're making it into a hybrid charm, and I think the behavior can be unexpected then or difficult to trace.

NohaIhab: Following this, we should add unit tests to verify the status reaches the unit.

NohaIhab (changes requested):
Thank you @mvlassis, I left some comments/questions

charms/kfp-profile-controller/src/charm.py
```diff
@@ -136,13 +139,36 @@ def __init__(self, *args):
...
+
+        self.s3_relation = self.charm_reconciler.add(
+            component=S3RequirerComponent(
+                charm=self,
+                name="relation:s3_credentials",
+                relation_name="s3-credentials",
+                is_optional=True,
+                required_relation_fields=frozenset({"access-key", "secret-key", "endpoint"}),
+            ),
+            depends_on=[self.leadership_gate, self.s3_relations_conflict_detector],
```
MattiaSarti, line 159: note: while depending on leadership is implicitly given by depending on the conflict detector, we can also leave it explicit, as it already is, to make it clearer at a glance

charms/kfp-profile-controller/src/charm.py
```diff
@@ -236,6 +225,140 @@ def __init__(self, *args):
...
+            MINIO_SECRET=json.dumps(
+                {"secret": {"name": f"{self.model.app.name}-minio-credentials"}}
+            ),
+            MINIO_HOST=object_storage["host"],
+            MINIO_PORT=object_storage["port"],
+            MINIO_NAMESPACE=object_storage["namespace"],
+            MINIO_ENDPOINT=object_storage["endpoint"],
+            MINIO_ACCESS_KEY=object_storage["access_key"],
+            MINIO_SECRET_KEY=object_storage["secret_key"],
+            MINIO_SSL="true" if object_storage["secure"] else "false",
```
MattiaSarti, line 270: suggestion (non-blocking): instead of turning a boolean into a string to then parse it back to a boolean, can't we directly keep a boolean in the first place? Or is this somehow more compatible with upstream?

charms/kfp-profile-controller/tests/integration/test_charm_object_storage.py
MattiaSarti, line 1: praise: well tested

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -144,7 +159,12 @@ def test_object_storage_relation_with_data(
 def test_object_storage_relation_without_data(
     harness: Harness, mocked_lightkube_client, mocked_kubernetes_service_patch
 ):
-    """Test that the object storage relation goes Blocked if no data is available."""
+    """Test that the object storage relation is Active when a relation exists but has no data.
+
+    The object-storage relation has minimum_related_applications=0, making it optional (a
+    relation with s3-credentials is an alternative), so even an empty relation does not block
+    the component.
+    """
```
MattiaSarti, line 167: suggestion (non-blocking): why was changing this necessary? The relation could still set the charm in a different status (be it blocked as before or waiting) if the relation is present but its data are not ready yet

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -202,7 +231,13 @@ def test_kubernetes_created_method(
     harness.charm.on.install.emit()
 
     # Assert
-    assert mocked_lightkube_client.apply.call_count == 4
+    # FIXME: This is a hardcoded count of the Kubernetes objects that should be created.
+    # `reconcile` is called 3 times (2 resources × 3 = 6):
+    #   - `object_storage_relation_changed` fires once but triggers 2 reconcile cycles because
+    #     both RelationCountGateComponent and SdiRelationDataReceiverComponent observe it
+    #     (CharmReconciler registers them independently without deduplication).
+    #   - `install` triggers a third reconcile cycle.
```
MattiaSarti, line 239: praise: essential comment

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -457,3 +508,182 @@ def test_policy_resource_manager_cleans_up_policies(
...
+
+@pytest.mark.parametrize(
+    "add_s3_credentials, add_object_storage, expected_status",
+    [
+        pytest.param(False, False, BlockedStatus, id="no-relation"),
+        pytest.param(True, False, ActiveStatus, id="s3-credentials-only"),
+        pytest.param(False, True, ActiveStatus, id="object-storage-only"),
+        pytest.param(True, True, BlockedStatus, id="both-relations"),
+    ],
+)
```
MattiaSarti, line 575: praise: very maintainable

charms/kfp-profile-controller/tests/unit/test_operator.py
```diff
@@ -457,3 +508,182 @@ def test_policy_resource_manager_cleans_up_policies(
...
+    """Test that an empty object-storage databag yields WaitingStatus."""
+    # Arrange
+    harness.begin()
+    harness.charm.object_storage_relation.component.get_data = MagicMock(return_value={})
+
+    # Act / Assert
+    with pytest.raises(ErrorWithStatus) as excinfo:
+        harness.charm._get_object_storage_data()
+    assert excinfo.value.status_type is WaitingStatus
+    assert excinfo.value.msg == "Waiting for object-storage relation data"
```
MattiaSarti: suggestion (non-blocking): let's assert directly the unit status, to also test that this propagates correctly through the rest of the code

MattiaSarti (approved):

charms/kfp-profile-controller/src/charm.py
MattiaSarti: along these lines: https://github.com/canonical/kfp-operators/pull/958#discussion_r3482614674)

charms/kfp-profile-controller/tests/integration/test_charm_s3.py
mvlassis: I'll add this in the last PR of the group, e.g. https://github.com/canonical/kfp-operators/pull/957

.github/workflows/ci.yaml
mvlassis: Addressed in https://github.com/canonical/kfp-operators/pull/958/commits/bcaa6213182844694c015c05115b129d8888e8ff (also changed the name of the `integration` environment in `kfp-api` to align)

charms/kfp-profile-controller/src/charm.py
mvlassis: Addressed in https://github.com/canonical/kfp-operators/pull/958/commits/bcaa6213182844694c015c05115b129d8888e8ff

charms/kfp-profile-controller/metadata.yaml
mvlassis: As explained in that comment, a CKF deployment with multiple relations of `kfp-api` should *not* be allowed. In the past, the charm was expecting a *single* integration, enforced by the `minium_related_applications` and `maximum_related_applications` in the `SdiRelationDataReceiverComponent` component, see https://github.com/canonical/kfp-operators/blob/0c3107e61f3e7758ff285a79a987a5bef6928891/charms/kfp-profile-controller/src/charm.py#L168

We now cannot rely on this, since we're making the specific component optional with the setting of `minimum_related_applications`

charms/kfp-profile-controller/src/charm.py
mvlassis: This is an interesting question, the point here is that we're dealing with environment variables here, and environment variables don't support a boolean time. At some point we *have* to make the conversion to a string, and [upstream does handle it similarly](https://github.com/kubeflow/pipelines/blob/845b490d8fdaa24a794e28a25def48052c83887d/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L108)

charms/kfp-profile-controller/tests/unit/test_operator.py
mvlassis: The change here was made because of the behavior in `SdiRelationDataReceiverComponent`. In the case when minimum_related_applications=1, `maximum_related_applications=1`, the status of the component is set to `Blocked` if the relation doesn't exist. This is *not* the behavior we actually want, because now the charm can operate when there is a relation `s3-credentials` and *not* `object-storage`.

The change essentially makes this component optional, which is what we want (optional on its own, but at least one of either `object-storage` and `s3-integrator`

charms/kfp-profile-controller/src/charm.py
mvlassis: The `get_object_storage` data is passed as a function argument in:
- the `KubernetesComponent` as `context_callable=`_generate_context`
- the `PebbleServiceComponent` as `inputs_getter=_generate_kfp_profile_controller`

Thus, these errors are actually called by the chisme reconciler, leading to the behavior we want.

Added unit test for that in https://github.com/canonical/kfp-operators/pull/958/commits/4f23c2a3730ce580e7321a998ad5257f41009bc9

charms/kfp-profile-controller/tests/unit/test_operator.py
mvlassis: Updated in https://github.com/canonical/kfp-operators/pull/958/commits/4f23c2a3730ce580e7321a998ad5257f41009bc9

charms/kfp-profile-controller/src/charm.py
MattiaSarti: Okay, then we'll leave it as is

charms/kfp-profile-controller/tests/unit/test_operator.py
MattiaSarti: Thanks for explaining, but I meant something slightly different by that: shouldn't the charm still be blocked (and so should this test still assert blocked status, as it used to) if the relation is present but has no data? Because in that case we know that the relation is the chosen one among the possible two ones, so it's not like when no relations are there, in which case it makes sense not to block, but when the expected relation is in place with no data we should have the charm complain about that via its status. Nevertheless, this is more than a thought and I'm fine with keeping this as is if it relies on Chisme and changes would be too time consuming.

MattiaSarti (approved):

charms/kfp-profile-controller/src/charm.py
NohaIhab: I prefer we do this with a dedicated component, for example like we're doing in the [`ConfigValidationComponent`](https://github.com/canonical/request-authentication-configurator/blob/main/src/components/config_validation.py) in the `request-authentication-configurator` charm, that way it's deterministic what status we'll set.

charms/kfp-profile-controller/tests/unit/test_operator.py
mvlassis: Moved to `BlockingStatus` in https://github.com/canonical/kfp-operators/pull/958/commits/db819c5bd9e5416439c5aecf3265446241924fa5

charms/kfp-profile-controller/src/components/object_storage_validator.py
```diff
@@ -0,0 +1,174 @@
...
+                # sync.py build a `host:port` endpoint (without a `.namespace` suffix).
+                "host": host,
+                "namespace": "",
+                "port": str(port),
+                "secure": secure,
+                "region": data.get("region", ""),
+                # The s3 interface has no namespace concept, so the endpoint is just host:port.
+                "endpoint": f"{host}:{port}",
+            }
+        data = active_storage_component.get_data()
```
NohaIhab: suggestion(non-blocking): can we put an `else` here to make it more readable that this only runs if we're not the s3 relation component

charms/kfp-profile-controller/src/components/object_storage_validator.py
```diff
@@ -0,0 +1,174 @@
...
+        # returns {} when no related app is present, otherwise a list of dicts. Extract the
+        # first (and expected-only) entry.
+        if isinstance(data, list):
+            data = data[0] if data else {}
+        if not data:
+            raise ErrorWithStatus("Missing object-storage relation data", BlockedStatus)
+
+        required_fields = ("access-key", "secret-key", "service", "namespace", "port")
+        missing = [f for f in required_fields if not data.get(f)]
+        if "secure" not in data:
```
NohaIhab: question: why is this not with the other `required_fields`? Is it not required? if so, then why do we always add it to the `missing` list

charms/kfp-profile-controller/src/components/object_storage_validator.py
NohaIhab, line 1: praise: neat design 🤩

NohaIhab (commented):
Thank you @mvlassis, just a Q on the `secure` field

charms/kfp-profile-controller/src/components/object_storage_validator.py
mvlassis: Should be addressed in https://github.com/canonical/kfp-operators/pull/958/commits/2a9982478b8b66d2336efa4a371361461a8d80ac (also added a comment)

charms/kfp-profile-controller/src/components/object_storage_validator.py
mvlassis: The check is correct, the issue is that the deserialization of the data converts the falsey "false" string into `False`, thus potentially returning a missing value incorrectly. See the [get_data](https://github.com/canonical/serialized-data-interface/blob/f45e1ea2ccc7b010c68f9336b16ba9189e3407c9/serialized_data_interface/sdi.py#L163) method in `serialized-data-interface`, and a [similar issue](https://github.com/containers/podman-compose/issues/1022#issue-2460757545)

NohaIhab (approved):
Thanks @mvlassis

PR #962, opened 2026-06-19
fix: handle multiple `istio-ingress-route` relations for `kfp-ui`

NohaIhab:
Part of https://github.com/canonical/bundle-kubeflow/issues/1446

Use `model.relations[...]` instead of `model.get_relation(...)` when checking ingress relations, so the charm no longer raises `TooManyRelatedAppsError` when more than one `istio-ingress-route` relation is present.

Add unit tests covering multiple `istio-ingress-route` relations and verifying each relation receives a valid `HTTPRoute` config.

_Note: this PR was created with the help of AI_


MattiaSarti (approved):
suggestion (non-blocking): as in other, analogous pull requests, we may add those clarifying comments before `self.ingress.submit_config()`

dariofaccin (approved):

